PR #3, opened 2020-03-17
Update README.md

deanhach:
On behalf of Daniel Kraft, change in title


deanhach:
After discussion a whole other set of edits.

PR #23, opened 2020-03-19
Cordova scaffold

myire:


myire:
going with ionic for v0.0.1.

PR #46, opened 2020-03-19
Add prettier action

theswerd:
CI now runs yarn prettier, then builds

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -27,6 +27,12 @@ jobs:
       - uses: actions/setup-node@v1
         with:
           node-version: '10.x'
+      - uses: creyD/prettier_action@v2.0
+        with:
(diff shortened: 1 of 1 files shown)


advayDev1:
I suggest that we do not do this with an action but instead as part of the pre-commit hook and via `yarn test`.  The action would just verify that the style is right and reject the PR if it is not (this is built into Sam's #31 once you add the prettier check on yarn test).  This is how @google/gts does it.

With an action, you are sending for review mis-styled code, so your colleagues have to deal with it in review, instead of preventing you from ever committing mis-styled code.

theswerd:
Sounds great, however I don't know how to do that. Do you have any links with instructions on how to do it?

advayDev1:
I would strongly suggest you wait for #33 to be submitted.  The earliest version of the app will be done with different framework and you'll need to integrate it into a different directory.

PR #72, opened 2020-03-20
[WHO-60] Add static intro about carousel and menu page

myire:


client/ionic/src/pages/About.module.css
```diff
@@ -0,0 +1,4 @@
+.page ion-header ion-toolbar {
+  --background: translucent;
+  --background: blue;
```
advayDev1: not sure what the effect of double specifying this property is?  does it mean anything?

client/ionic/src/pages/Menu.module.css
```diff
@@ -0,0 +1,4 @@
+.page ion-header ion-toolbar {
+  --background: translucent;
+  --background: blue;
```
advayDev1: ditto

myiremark: removed in https://github.com/WorldHealthOrganization/app/pull/74

client/ionic/src/pages/About.module.css
myiremark: removed in https://github.com/WorldHealthOrganization/app/pull/74

myiremark:
Created screenshots, we need to figure out where to put them and are we adding to docs.

Closing this PR in favor of https://github.com/WorldHealthOrganization/app/pull/74 

also, how do we want to handle merges of upstream?  That may just have been from me being an idiot last night

PR #111, opened 2020-03-21
Fix card subtitle spelling

wdibi:
Fixed two incorrectly spelled words.


pjayathissa:
I've taken these spelling mistakes in the upcoming css PR

PR #121, opened 2020-03-21
Optimize CI process with artifacts

epicfaace:
for #119

I didn't reinstall node_modules because I think that may be OS-dependent. Instead, I uploaded the result from `ionic build` into a build artifact and then uploaded it.

Also, changed it to use `npm ci` instead of `npm install` because this is [much faster](https://docs.npmjs.com/cli/ci.html) than npm install and suitable for CI process.

### DIFF
--- a/.github/workflows/client.yml
+++ b/.github/workflows/client.yml
@@ -12,7 +12,7 @@ on:
       - '.github/workflows/**'
       - 'client/ionic/**'
 jobs:
-  test:
+  test-and-build:
     strategy:
       matrix:
         platform: [ubuntu-latest, macos-latest]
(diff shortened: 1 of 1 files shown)


.github/workflows/client.yml
```diff
@@ -48,13 +52,15 @@ jobs:
...
-        working-directory: ./client/ionic
-      - run: ionic build
+      - run: npm ci
         working-directory: ./client/ionic
-      - run: ionic cap sync
+      - uses: actions/download-artifact@v1
+        with:
+          name: build
+          path: ./client/ionic/build
+      - run: npx capacitor sync
```
epicfaace: `ionic cap sync` runs both `ionic build` and `capacitor sync`. We've already gotten the results of the former through the former so we only need to run the latter.

epicfaace:
I'll break this into smaller PRs.

PR #127, opened 2020-03-21
Set up App Engine.

crazybob:
This sets up App Engine, Cloud Datastore, and Protocol Buffer-based APIs. It currently deploys to https://who-app-staging.appspot.com/.


client/rpc/generate-api.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+# Generates Javascript stubs for our API
+
+cd $(dirname "$0")/..
+
+# TODO: Update output directory
+java -jar lib/present-rpc-compiler.jar --proto_path=../proto --js_out=react/src
```
advayDev1: What's the versioning, license, etc. story with this JAR?  Do we need the binary checked in directly?

crazybob: This is my project. It's Apache: https://github.com/presentco/present-rpc

I went with checking in a jar to simplify dev environment config. The jar is really small.

advayDev1: Since we're copying the binary in to the repo can we 1) add a comment here of exactly how to reproduce it (I assume just https://github.com/presentco/present-rpc and a particular git sha), and 2) add to THIRD_PARTY_LICENSE your license details.

server/appengine/build.gradle
```diff
@@ -0,0 +1,91 @@
...
+    sourceCompatibility = 1.8
+    targetCompatibility = 1.8
+}
+
+configurations.all {
+    // This is pulled in via appengine-gcs-client. :-(
+    exclude group: 'com.google.guava', module: 'guava-jdk5'
+}
+
+dependencies {
```
advayDev1: Per CONTRIBUTING.md please describe the dependencies and justifications.  I recognize all of the deps as widely-used (personally) except your present RPCs, roboslack, and objectify.

advayDev1:
Sweet - just left 2 comments regarding dependencies.

server/appengine/build.gradle
crazybob: Present RPC: It's just protocol buffers/JSON over HTTP. There's isn't a widely-adopted alternative that I know of. (GRPC requires HTTP/2, isn't suitable for use from web apps, etc.).

Roboslack: Just Java objects modeling Slack's API.

Objectify: The best way to access Cloud Datastore from Java. Google's own docs cover it extensively.

client/rpc/generate-api.sh
crazybob: Linked to download instructions and updated license file.

BTW, this tool just generates Javascript functions from proto rpcs. Easy enough to code these by hand, too.

advayDev1:
Please add your name into LICENSE.  LGTM.

We have a lot more leeway on changing deps on server-side so I'm a less paranoid about it than client ones.

advayDev1:
I believe our latest rules say 2 LGTMs - please pick your second reviewers and assign the PR to them.

PR #136, opened 2020-03-21
WHO Protect Yourself Page

myiremark:


myiremark:
00_wash_hands
01_cover
02_closed_bin
03_wash_hands_often
04_avoid_close_contact
05_seek_medical_care

client/ionic/src/App.tsx
```diff
@@ -32,6 +33,11 @@ const App: React.FC = () => (
       <IonRouterOutlet>
         <Route path="/about" component={About} exact={true} />
         <Route path="/menu" component={Menu} exact={true} />
+        <Route
+          path="/protect_yourself"
```
hspinks: recognizing that this is totally a personal preference thing, but I prefer `kebab-case` to `snake_case` in URL paths. open to discussion though

myiremark: i'd say it matters.  WHO uses kebab-case on their website.  I don't have a preference, just python baked into my brain

advayDev1: these paths can't be seen by the user (at least if we're not a true webapp) so meh. but  i like kebabs too (and kebab case ;)

myiremark: intention is to get scaffolds in for all pages.  closing this pr, 

reopened as https://github.com/WorldHealthOrganization/app/pull/141

for branch name, also, switching to dynamic content.

myiremark: will also update link in slack

PR #152, opened 2020-03-22
Make UI respect safe space

### LINKED ISSUE #143: UI must respect device 'safe space'
Different devices have different safe spaces (ex. some models have cutouts of the top screen.)

See below how this affects iPhone X

ReactNative has https://reactnative.dev/docs/safeareaview - quick google did not show me an equivalent Ionic API.

Screen Shot 2020-03-21 at 6 38 14 PM

darcien:
A quick fix.
~This only works for portrait orientation. In the future, we might need to handle other orientation.~

image

This will fix #143


SamMousa:
Let's do this for all sides: https://css-tricks.com/the-notch-and-css/

darcien:
> 
> 
> Let's do this for all sides: [css-tricks.com/the-notch-and-css](https://css-tricks.com/the-notch-and-css/)

Done!

darcien:
> 
> 
> Please add all padding variables for safe space.

Padding doesn't work, that's why I used margin instead.

SamMousa:
It seems there are theme variables available for this; also it should work out of the box. Could you have a look at this https://ionicframework.com/docs/theming/advanced?

If it doesn't work we might have just misconfigured something. We should be able to reconfigure, or worst case manually initialize those variables with the values you used in your css proposal.

darcien:
> It seems there are theme variables available for this; also it should work out of the box. Could you have a look at this https://ionicframework.com/docs/theming/advanced?

I looked into it, but seems like the value provided by ionic is the same with the env variable, e.g. (env(safe-area-inset-top) === var(--ion-safe-area-top)).

Which means it's already initialized but not used.

I'm not really familiar with the Ionic environment, but I think this should be handled by ionic.

There's a similar issue on capacitor repo:
image
https://github.com/ionic-team/capacitor/issues/777#issuecomment-424603988

But I'm not sure about injecting some stylesheet from the network, especially when we might need offline support later on.

SamMousa:
We should have that locally available in node modules folder.

darcien:
Ah right, I updated the PR to use those variables, could you please check again, thanks.

PR #168, opened 2020-03-22
Treat eslint warnings as errors on build

epicfaace:
By setting `CI=true` during `ionic build`, we can make sure that eslint warnings are treated as errors on build. This PR sets that setting and also fixes existing eslint errors (`array-callback-return`, `@typescript-eslint/no-unused-vars`).

### DIFF
--- a/.github/workflows/client.yml
+++ b/.github/workflows/client.yml
@@ -52,6 +52,8 @@ jobs:
         working-directory: ./client/ionic
       - run: ionic build
         working-directory: ./client/ionic
+        env:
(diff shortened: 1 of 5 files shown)


epicfaace:
Actually, there's a problem here. It appears that react-scripts has its own eslint config, but gts has another eslint config. I'd assume we want to switch over to gts's eslint config -- but react-scripts doesn't let you override the eslint config used.

Probably the best way forward is just not to merge this PR and ignore all the warnings from react-scripts that come when running the app.

PR #214, opened 2020-03-23
Allow usage of markdown in bodyTexts

epicfaace:
Allows usage of markdown in bodyTexts. This is important because in certain pages we want to display not just plain text, but formatting. Markdown is sufficient for basic formatting we need such as bolding / italics.

I intentionally decided against rendering HTML in order to reduce complexity of what can be added in the `.yml` file -- and by allowing too complex HTML, it will be difficult to maintain multiple copies of the yml file for translations.

Added react-markdown (https://github.com/rexxars/react-markdown) to render markdown (MIT license, 5k stars on GitHub)

I've only added bolding to the "Protect Yourself" yml file because that seems to be the only one with the right content as of now. You can see that the bolding works now:

image


epicfaace:
We'll do a HTML parser that only allows `<em>` instead.

PR #215, opened 2020-03-23
Allow usage of html in bodyTexts

epicfaace:
Allows usage of HTML in bodyTexts. This is important because in certain pages we want to display not just plain text, but formatting. Right now, we sanitize the HTML and only allow the `<em>` tag.

Added DOMPurify (https://github.com/cure53/DOMPurify) to sanitize HTML and restrict allowed tags (Apache License Version 2.0, 5k stars on GitHub).

Also added @types/DOMPurify (MIT license, see https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/LICENSE)

I've only added the `<em>` changes to the "Protect Yourself" yml file because that seems to be the only one with the right content as of now. You can see that the emphasis works now (we can later make it bold in a future styling update):

image


advayDev1:
Didn't you agree with @SamMousa just now to use HTML, with limited tags, not markdown?

epicfaace:
@advayDev1 oops, didn't push my commits. The PR should be updated now.

client/ionic/package.json
```diff
@@ -17,6 +17,7 @@
     "@types/react-dom": "^16.9.4",
     "@types/react-router": "^5.1.4",
     "@types/react-router-dom": "^5.1.3",
+    "dompurify": "^2.0.8",
```
advayDev1: File an issue to review security settings on dompurify.  https://snyk.io/vuln/npm:dompurify

PR #217, opened 2020-03-23
Improve carousel styling

epicfaace:
- Refactor each page with a carousel to use a common Carousel component
- Improve Carousel styling (color of pagination, font size / placing) to match the Figma mockup

for #215

No

image

image


client/ionic/src/components/Carousel.tsx
```diff
@@ -0,0 +1,74 @@
...
+}: {
+  screen: TextImageScreen;
+  flow: LoadedFlow;
+}) => {
+  return (
+    <IonSlide>
+      <div className="flex flex-column justify-around h-100 pa4">
+        <div className="mt3 tl">
+          <div style={{ fontSize: '1.5em' }}>{screen.headingText}</div>
+          <div style={{ fontSize: screen.headingText ? '1em' : '1.5em' }}>
```
hspinks: clarifying - when there is a `headingText` in the screen, the font size of the body text is smaller?

client/ionic/src/components/Carousel.tsx
```diff
@@ -0,0 +1,74 @@
...
+  flow,
+}: {
+  screen: TextImageScreen;
+  flow: LoadedFlow;
+}) => {
+  return (
+    <IonSlide>
+      <div className="flex flex-column justify-around h-100 pa4">
+        <div className="mt3 tl">
+          <div style={{ fontSize: '1.5em' }}>{screen.headingText}</div>
```
hspinks: nit: recommend conditionally rendering this div based on whether `headingText` exists on the screen rather than leaving an empty div

advayDev1:
@epicfaace needs conflict res.  do you need the import tachyons on the pages now?

tested locally, seems fine except the font...

client/ionic/src/components/Carousel.tsx
epicfaace: Yes. Initially, I was just going to make `headingText` equal to 1em as in the below image:

image

However, it appears that the splash carousel content also has a header and a long body (this wasn't in the mocks, but I decided to leave it as is) -- and changing the body text to be smaller made it look a lot better:

image

client/ionic/src/components/Carousel.tsx
epicfaace: Done

epicfaace:
> @epicfaace needs conflict res. do you need the import tachyons on the pages now?
> 
> tested locally, seems fine except the font...

You're right, tachyons doesn't need to be imported. Fixed that.

Font should work on mobile -- it's probably the case that your computer doesn't have Roboto pre-installed. We should probably import Roboto through Google Web Fonts whenever Ionic is running on the web, but that's for another issue / PR...

client/ionic/src/components/Carousel.tsx
hspinks: sounds good, thanks for clarifying! marking this as resolved

hspinks:
On my iPhone simulator, I'm not seeing the progress indicator bubbles on the bottom of the carousel:

Screen Shot 2020-03-23 at 7 46 25 PM

Even though I am seeing them when running the app in the browser:

Screen Shot 2020-03-23 at 7 48 57 PM

Can you take a look?

PR #233, opened 2020-03-24
feat: TextEmoji screen and WHO Myth-busters page

### LINKED ISSUE #227: Create TextEmoji screen type
The information we have on the chat bot displays information cleanly with an associated emoji. I figure that could be a decent way to handle the issues of imagery. Create a `TextEmoji` screen type to handle this.

hspinks:

Creates a new TextEmoji screen type that renders text at the top of the screen and a large emoji (as text) at the bottom. Creates a new WHO Myth-busters page using TextEmoji screens, and updates the Protect Yourself page to use TextEmoji screens.

I believe this solves 2 problems for us:
1. Approved, license-compatible, relevant imagery has been hard to come by. I believe we should be able to use emojis in the app without worrying about licenses, since they are rendered by the OS
2. We are attempting to build our content to match the WHO WhatsApp Bot functionality, and the bot already uses these emoji - makes it easy to transfer the content to our app

Potential concerns to investigate with using emoji:
* Will there be any cross-device rendering issues with emoji?
* Will we be losing support for some older devices and/or operating systems that can't render emoji?
  * Including if any emoji we are using were added in a later spec
* Are there any accessibility concerns with using emoji?

I think that the above concerns are somewhat mitigated by the fact that the emoji is just auxiliary content on the page, and it is the text that is the essential information of the screen. So someone can still get the primary meaning of each screen even if the emoji doesn't render exactly as we hope.

Closes #227 

N/A

* [x] In browser
* [x] On iOS Simulator

### DIFF
--- a/client/ionic/src/App.tsx
+++ b/client/ionic/src/App.tsx
@@ -10,6 +10,7 @@ import SplashInfo from './pages/SplashInfo';
 import SplashCarousel from './pages/SplashCarousel';
 import Menu from './pages/Menu';
 import ProtectYourself from './pages/ProtectYourself';
+import WhoMythbusters from './pages/WhoMythbusters';
 
 /* Core CSS required for Ionic components to work properly */
 import '@ionic/react/css/core.css';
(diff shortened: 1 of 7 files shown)


hspinks:
Screenshots...

TextEmoji screen on iOS Simulator:
Screen Shot 2020-03-24 at 1 07 39 AM

TextEmoji screen on Browser:
Screen Shot 2020-03-24 at 1 08 08 AM

Main menu on iOS Simulator:
Screen Shot 2020-03-24 at 1 09 48 AM

SamMousa:
image
Works on android.

Issue #218, opened 2020-03-23
Change wording from "World Health Organization App"


epicfaace:
Right now, the description for the repo says "World Health Organization App" and the README says "This is the official World Health Organization app." We should make it clear that this is not the official app for the WHO, but rather a WHO app focused on the coronavirus pandemic.

ctrimm:
Made an attempt at updating the text in the README.md. Welcome any and all feedback on that open PR.

advayDev1:
This is still something very much in flux right now and there aren't decisions.

PR #253, opened 2020-03-24
Updated README.md Verbiage to Reflect Feedback

ctrimm:
Updates README.md to be more specific around the application's description per issue #218 

N/A

Reviewed locally & on Github.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -1,6 +1,6 @@
 # World Health Organization Coronavirus Pandemic App
 
-This is the official World Health Organization application focused on the Coronavirus Pandemic.
+This is the official World Health Organization application focused on the coronavirus pandemic.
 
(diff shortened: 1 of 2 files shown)


docs/README.md
```diff
@@ -1,6 +1,6 @@
-# World Health Organization App
+# World Health Organization Coronavirus Pandemic App
 
-This is the official World Health Organization app. 
+This is the official World Health Organization application focused on the Coronavirus Pandemic.
```
epicfaace:
```suggestion
This is the official World Health Organization application focused on the coronavirus pandemic.
```

ctrimm:
Updated per your feedback @epicfaace. It will need another approval.

hspinks:
Definitely appreciate that we're going to need to clarify some of the language around this, but to echo the [Advay's comment](https://github.com/WorldHealthOrganization/app/issues/218#issuecomment-604187930), I'm a bit wary of pushing out new messaging while there are still some directional decisions in process. I'd prefer to update that communication once when we have that direction

PR #256, opened 2020-03-24
fix: Add Icons and Splash

sstur:

This adds the icons and splash screen, replacing the default Ionic assets.

Note: this is a re-hash of #189, but it still doesn't contain the symlinks or the new artwork, so I'm marking as a draft until we get that done.

No.

Opened in Xcode and Android Studio.

Screen Shot 2020-03-22 at 15 52 30
Screen Shot 2020-03-22 at 15 56 16
Screen Shot 2020-03-22 at 15 57 08


kassim:
the icon doesn't render correctly for Android

Screenshot_1585095769

I've fixed it here https://github.com/sstur/who-app/pull/1

advayDev1:

Issue #56, opened 2020-03-20
Review app required permissions in AndroidManifest.xml


patniemeyer:
Review the permissions requested in: client/ionic/android/app/src/main/AndroidManifest.xml

RohanTalip:
Does this need to be in the v0.1 milestone?

myiremark:
@crazybob do we need any of the permissions in:

https://github.com/WorldHealthOrganization/app/blob/master/client/ionic/android/app/src/main/AndroidManifest.xml?

myiremark:
@shankari

brunobowden:
we must remove all unnecessary permissions for the first public release. We can always add them back selectively later on

PR #273, opened 2020-03-24
Removed some currently unnecessary permissions from AndroidManifest.xml

RohanTalip:

It removes some currently unnecessary permissions from AndroidManifest.xml

See issue #56 

No.

I built the app locally, and ran it in the Android emulator.

### DIFF
--- a/client/ionic/android/app/src/main/AndroidManifest.xml
+++ b/client/ionic/android/app/src/main/AndroidManifest.xml
@@ -42,12 +42,15 @@
         </provider>
     </application>
 
-    <!-- TODO: Review requires permissions: https://github.com/WorldHealthOrganization/app/issues/56 -->
     <!-- Permissions -->
 
     <uses-permission android:name="android.permission.INTERNET" />
+
     <!-- Geolocation API -->
+    <!-- TODO: uncomment when location permissions are actually needed. -->
+    <!--
     <uses-permission android:name="android.permission.ACCESS_COARSE_LOCATION" />
     <uses-permission android:name="android.permission.ACCESS_FINE_LOCATION" />
     <uses-feature android:name="android.hardware.location.gps" />
+    -->
 </manifest>
--- a/client/ionic/android/app/src/main/AndroidManifest.xml
+++ b/client/ionic/android/app/src/main/AndroidManifest.xml
@@ -46,19 +46,8 @@
     <!-- Permissions -->
 
     <uses-permission android:name="android.permission.INTERNET" />
-    <!-- Camera, Photos, input file -->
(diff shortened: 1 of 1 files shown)


RohanTalip:
I left the location permissions in the file for now, as it was less clear that these need to be removed, given eventually the app may want to access the device's location.

client/ionic/android/app/src/main/AndroidManifest.xml
```diff
@@ -46,19 +46,8 @@
...
     <uses-permission android:name="android.permission.INTERNET" />
-    <!-- Camera, Photos, input file -->
-    <uses-permission android:name="android.permission.READ_EXTERNAL_STORAGE"/>
-    <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE" />
     <!-- Geolocation API -->
     <uses-permission android:name="android.permission.ACCESS_COARSE_LOCATION" />
     <uses-permission android:name="android.permission.ACCESS_FINE_LOCATION" />
     <uses-feature android:name="android.hardware.location.gps" />
-    <!-- Network API -->
-    <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
```
epicfaace: Why don't we need ACCESS_NETWORK_STATE?

myiremark: Also, for the first version, do we actually need access location?

shankari: @epicfaace `ACCESS_NETWORK_STATE` allows you to control whether to access the internet only when connected to WiFi or sth like that. Typically not used for foreground operation, and v0.1 app is all foreground

@myiremark I believe the answer at the standup meeting was "no"

brunobowden:
@RohanTalip - thanks for cleaning this up

client/ionic/android/app/src/main/AndroidManifest.xml
RohanTalip: My understanding of ACCESS_NETWORK_STATE is that it is needed if you want to check if you are connected to a network or not, i.e. not likely needed at present.  The Ionic code I looked at today only seems to use it in the [Network API](https://github.com/ionic-team/capacitor/blob/master/site/docs-md/apis/network/index.md).  The other reference I found was in the [(Ionic) Capacitor Android Plugin Guide
](https://github.com/ionic-team/capacitor/blob/master/site/docs-md/plugins/android.md).

I'll comment out the location access for now in another commit, coming shortly.

RohanTalip: The location access permissions have been commented out for now.

advayDev1:

PR #284, opened 2020-03-25
Merge Flutter client from original repo...

patniemeyer:


advayDev1:
- [ ] Remove PDF and asset pngs except logo
- [ ] Waiting on confirmation of how page_view_indicator is pulled in by @theswerd

client/flutter/lib/home_page.dart
```diff
@@ -0,0 +1,49 @@
...
+                .push(MaterialPageRoute(builder: (c) => WhoMythBusters()))),
+        PageButton(
+            title: S.of(context).travelAdvice,
+            onPressed: () => Navigator.of(context)
+                .push(MaterialPageRoute(builder: (c) => TravelAdvice()))),
+        PageButton(
+            title: S.of(context).shareTheApp,
+            lightColor: true,
+            onPressed: () => Share.share(
+                'Check out the official COVID-19 GUIDE App https://preview.whoapp.org/menu')),
```
advayDev1: Not blocking check-in this has to go away

patniemeyer: hm.  easily changed if you have a suggestion for tonight...

advayDev1: That URL would be a url to the app itself on the appropriate app store.

We have no idea what the bundle id is for Apple yet for example.

client/flutter/android/app/src/main/kotlin/com/example/WHOFlutter/MainActivity.kt
```diff
@@ -0,0 +1,12 @@
+package com.example.WHOFlutter
```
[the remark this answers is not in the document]
patniemeyer: So this automatically generated hook is where the Flutter plugins register their native components.  More generally this is the place where we might end up adding a small amount of native code in the future if needed.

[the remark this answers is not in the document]
patniemeyer: Ah, I see the com.example domain was mistakenly left in there making it appear that this is example code.  I will fix.

PR #285, opened 2020-03-25
Update LICENSE for other WHO brand assets

advayDev1:
Just to make it easier to commit Pat's PR.


advayDev1:
Continuing based on Slack ack.

PR #288, opened 2020-03-25
Tap to move through slides

theswerd:
This will make it so when a user taps on a slide, that it moves to the next one.

Nope

iOS simulator

### DIFF
--- a/client/flutter/lib/carousel_page.dart
+++ b/client/flutter/lib/carousel_page.dart
@@ -48,12 +48,17 @@ class CarouselView extends StatelessWidget {
 
   final pageIndexNotifier = ValueNotifier<int>(0);
 
+  PageController pageController = new PageController();
+
   @override
   Widget build(BuildContext context) {
     return PageScaffold(
+      
       body: Stack(
         children: <Widget>[
+
           PageView(
+            controller: pageController,
             onPageChanged: (i) => pageIndexNotifier.value = i,
             children: this.items,
           ),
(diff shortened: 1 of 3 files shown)


patniemeyer:
@theswerd This is good but the elastic curve causes the card to do a little dance before it starts moving.  I suggest `curve: Curves.easeOut` maybe?  Also when you reach the last card we should ignore the tap (currently pushes the card off the screen briefly).  Other than those looks great!

patniemeyer:
@theswerd Maybe `Curve.easeInOutCubic` is better... a bit faster.

theswerd:
@patniemeyer done!

client/flutter/lib/carousel_page.dart
```diff
@@ -63,6 +68,10 @@ class CarouselView extends StatelessWidget {
                 padding: EdgeInsets.only(bottom: 20),
                 child: pageViewIndicator(context)),
           ),
+          GestureDetector(
+          onTap: ()=>pageController.page == this.items.length-1?null:pageController.nextPage(duration: Duration(milliseconds: 500), curve: Curves.easeInOutCubic)
```
sstur: Nit: Is the spacing around operators here intentional? Should we incorporate dartfmt in our pipeline to standardize style?

theswerd: When I wrote this my IDE wasn't working, I will re push after formatting

theswerd: @sstur thx, just fixed

creativecreatorormaybenot: @sstur Check #362.

PR #289, opened 2020-03-25
Speculative - Add Cocoapod dependencies to git.

patniemeyer:
Adding this in case it is needed.


hspinks:
Noting that the BUILD docs in #297 assume that CocoaPods dependencies will need to be installed, meaning they won't need to be checked into the repo if those instructions work

epicfaace:
@patniemeyer what are the main benefits of doing this? (given that we already have the `Podfile`and `Podfile.lock` checked in)

hspinks:
@epicfaace this was a "just in case" PR opened to have it ready just in case we needed it for deployment

@patniemeyer since we got over this hurdle without this and will hopefully move to more automatic delivery going forward, I'm going to close this. thanks for being on top of preparing for every eventuality! we can always re-open if we need

PR #316, opened 2020-03-25
Updated localisation files with MT

clementmouchet:
Applied MT to the localisation files for the flutter client


clementmouchet:
@patniemeyer / @theswerd feel free to take this or reject it, it's just MT to see how that renders in the app

hspinks:
thanks @clementmouchet! this will be helpful for testing I think. one thing that was mentioned in standup today, not sure if there was a decision made on it... should we mark any machine translation with a message indicating it was translated automatically?

clementmouchet:
So I think what matters more it to agree on a lifecycle, where MT is used and refreshed periodically while the app is being worked on and as we get close to release, we freeze adding/editing the arb files, send them for professional translation and override the content in the repo with the proper files for the release.

As discussed with @patniemeyer & @theswerd the content base is really small now, and it's management isn't an issue. Going forward, as content gets authored, a strategy should be taken to manage this differently. I don't think localisation files will work for that. They are great for the static content in the app and short dynamic content with variable placeholders, plural forms, genders etc.

Beyond that I think there's a need for a proper CMS+API that supports the authoring and localisation of content. A CMS would allow great autonomy and reactivity from content authors, and one with revision & localisation support would allow editing content without blocking people from making edits while content is being translated etc.

That's just food for thoughts, the point of this PR & MT is just to raise awareness of how other languages appear in the app.

PR #317, opened 2020-03-25
Fix overflow issues on carousel slider

### LINKED ISSUE #303: UI Overlap on large text
there's some UI overlap of text overlapping the pager indicator when checking on a Nexus 5 emulator

it doesn't overlap on iPhone 8 but there is a margin issue, I'm not sure if that will block release

<details><summary>Android Nexus 5 (1080x1920 pixels) - click to expand</summary>

Screenshot_1585136324

Screenshot_1585136488

</details>

<details><summary>iPhone 8 simulator - click to expand</summary>

Simulator Screen Shot - iPhone 8 - 2020-03-25 at 12 45 39

</details>

dnfield:

Fixes #303 

This wraps the content of CarouselSlider pages in a `SingleChildScrollView`, so that if the text would otherwise overflow on the screen it can be scrolled into view.

It also gives an elevation to the page view indicator dots to make them visually more distinct from text that might be flowing under them.

Eliminates the `context` argument on the widget since that can cause confusion with the actual context you want to use in the build function.

Uses some sizedBoxes to normalize the way that spacing is handled, so that the text doesn't appear to jump up and down from page to page when pages have different amounts of text (and, in particular on iOS, so that Emoji font resolution doesn't get silly when it thinks one emoji is bigger than another).

Applied some formatting.

No.

Manually, as well as writing widget tests that will fail if long text overflows and test that we can scroll hidden text into view.


client/flutter/lib/carousel_page.dart
```diff
@@ -81,35 +86,38 @@ class CarouselView extends StatelessWidget {
...
-                size: 8.0,
-                color: Colors.grey,
-              ),
-              highlightedBuilder: (animationController, index) =>
-                  ScaleTransition(
-                scale: CurvedAnimation(
-                  parent: animationController,
-                  curve: Curves.ease,
+    return Material(
+      elevation: 4.0,
```
hspinks: do we want to have a standardized system for elevation constants, or overkill at this point?

dnfield: 4.0 is pretty arbitrary here - if there's a better place to capture this I'm fine with that.

gspencergoog:
Oh, @dnfield, looks like I might have done a similar thing in my PR when addressing the text scale factor isssues: https://github.com/WorldHealthOrganization/app/pull/319

Sorry, I should have coordinated with you.

gspencergoog:
I'm happy to resolve conflicts with my changes after you land.

dnfield:
No worries @gspencergoog!  Anything you think I should change here given what you were working on? I didn't do as much testing with different text scale factors.

advayDev1:
Don't forget to put your name in content/credits.yaml :)

dnfield:
@advayDev1 done

client/flutter/test/widget_test.dart
```diff
@@ -5,26 +5,13 @@
 // gestures. You can also use WidgetTester to find child widgets in the widget
```
advayDev1: can remove this comment now

dnfield: Done

client/flutter/lib/carousel_page.dart
hspinks: fine with this for now!

patniemeyer:
@dnfield Thanks for looking into this.  A couple of comments - 

1) I think we should probably get some input from the designers before we do to much with the page indicator.  It does look cool but adding elevation really changes the feel of the app and is a bit contrary to current trends where the designers like the "flat" look.

2) Looks like you used a good approach to the sizing issue however the layout is no longer utilizing the space on larger devices. (both the image and the text sort of sit at the top of the screen).  We could improve that by adding a `Spacer()` to the top to absorb some of the unused space (you'll see what I mean if  you try adding one as the first element of your Column) , however then the image and text both change position as you scroll through, which isn't ideal.

I think we may want to allocate a fixed portion of the screen to the graphic and to the text respectively and then allow the text to scroll if needed... We can also do a media query and adjust the font scale on smaller devices as a general workaround, which may mean that it never actually has to scroll (We do this for the main menu currently).

I generally test on the iPhone Max Pro and the iPhone SE sizes to hit the two extremes and those sizes are represented in that simple scaling logic I mentioned above.

The above said - if everyone is happy with the design changes I am fine with them... The code looks good.  Thanks for diving in!

advayDev1:
I understand your concerns @patniemeyer - I think our design is so up in the air right now, that its not a big deal right now. SG?

dnfield:
@patniemeyer -

I'll drop the `Material` addition and elevation for now, and filed https://github.com/WorldHealthOrganization/app/issues/325 to figure out a way to resolve that part.

I've updated the PR a bit as well to use more of the vertical space, basing it on the height of the scrollable area using a LayoutBuilder.  PTAL.

client/flutter/pubspec.yaml
```diff
@@ -4,7 +4,7 @@ description: WHO Flutter app
 version: 0.1.0+1
 
 environment:
-  sdk: ">=2.1.0 <3.0.0"
+  sdk: ">=2.3.0 <3.0.0"
```
myiremark: I'm not familiar enough with Dart sdk versions.  What impact does this have on device support (if any)?

patniemeyer:
I think we probably have overlap here with PR #319 .  Sorry if my delay in reviewing is causing duplication of effort.

client/flutter/pubspec.yaml
dnfield: None.  This is to allow use of newer language features.

dnfield:
@gspencergoog and I can work to resolve merge conflicts.  There's some commonality between these two.  If #319 covers things better we can go with that one.

gspencergoog:
I'm happy to resolve any merge conflicts in my PR after this one lands.

client/flutter/pubspec.yaml
akeblom: You should be able to bump it up to 2.7.0 if you run the latest flutter version on the stable channel.

dnfield: The hotfix containing that was just recently release, but yes that should be possible.  It isn't needed for this patch specifically though.

dnfield:
Fixed up merge conflicts.  Is this good to land?

patniemeyer:
@dnfield Thank you again for your work.  Are there things in this PR that weren't addressed by #319 ?

dnfield:
This keeps the UI elements at consistent heights as you swipe through, and bases those heights on the size of the viewport of the scrollable column rather than on the size of the whole screen.

#319 also adds some scroll down hinting and some other text layout fixes for various text scaling values.

dnfield:
The UX of this has all changed substantially.  I'm going to close this, since it's no longer relevant.

Issue #325, opened 2020-03-26
Carousel progress indicator is difficult to distinguish from text if text is long


dnfield:
On a smaller phone screen, some of the longer text information can end up overflowing into the page view dots.  It becomes difficult to tell that there's more text to scroll down to view.

One way to solve this would be to add an elevation to the progress dots. Another way might be to put them behind a gradient or a frost effect into the text (effectively making the text slide under them).

KhaleelShaheenDC:
I worked on this issue at this PR:
https://github.com/WorldHealthOrganization/app/pull/360

samueladekunle:
Hello @dnfield, as your PR on this issue been accepted?

dnfield:
I removed the code for this from my pr until a design is settled on

hspinks:
Closing as design has changed

Issue #320, opened 2020-03-26
Carousel page indicator overflow


dnfield:
The carousel dots at the bottom of the screen do not scale to handle large numbers of pages.

This is confusing when there are additional facts to view but the dots seem to indicate you are at the end.

This screen shows the indicator at the last dot, even though there are more pages:

![flutter_01](https://user-images.githubusercontent.com/8620741/77598194-683ea900-6ebe-11ea-85c9-3f04f6cf9655.png)

Here's what it looks like when you keep swiping to the new page:

![flutter_02](https://user-images.githubusercontent.com/8620741/77598206-6d9bf380-6ebe-11ea-9246-6b9676eeb789.png)

richikchanda1999:
I have solved this issue but I am not a part of the project as of now.
Should I send a PR or is there a way I can contribute as well?

I have already filled the Google Form.

hspinks:
Closing as design has changed

PR #360, opened 2020-03-26
:iphone: refactor(carousel_page): use auto scrolling carousel indicator

KhaleelSH:
Remove `page_view_indicator` package and use `scrolling_page_indicator` instead. This is a better user experience when there is many pages than the screen width can show.

Adds auto scrolling carousel.

This PR solves the issue when there is lot's of pages, and hence page indicator circles, than the screen width can show.

Now, it shows an auto scrolling page progress indicator, similar to Instagram's carousel indicator

Issues affected:
#320 
#325 

I removed `page_view_indicator` package.
I added `scrolling_page_indicator`

Tested on Android and iOS emulators.
Tested on Android device.

image
image


advayDev1:
Thanks for your interest.  This carousel is no longer in the current UI.

PR #361, opened 2020-03-26
Fix Dart SDK constraint

creativecreatorormaybenot:

The Dart SDK constraint should match Flutter **stable**. There is no benefit in not using the minimum constraint from stable.

With this PR, we will be able to use the latest stable Dart SDK features.

**No**.

Ran all existing tests.

### DIFF
--- a/client/flutter/.gitignore
+++ b/client/flutter/.gitignore
@@ -36,3 +36,6 @@ lib/generated_plugin_registrant.dart
 /coverage/
 # Exceptions to above rules.
 !/packages/flutter_tools/test/data/dart_dependencies_test/**/.packages
+
+# Clean commits
+pubspec.lock
--- a/client/flutter/.gitignore
+++ b/client/flutter/.gitignore
@@ -36,6 +36,3 @@ lib/generated_plugin_registrant.dart
 /coverage/
 # Exceptions to above rules.
 !/packages/flutter_tools/test/data/dart_dependencies_test/**/.packages
-
-# Clean commits
-pubspec.lock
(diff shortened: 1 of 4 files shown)


client/flutter/.gitignore
```diff
@@ -36,3 +36,6 @@ lib/generated_plugin_registrant.dart
 /coverage/
 # Exceptions to above rules.
 !/packages/flutter_tools/test/data/dart_dependencies_test/**/.packages
+
+# Clean commits
+pubspec.lock
```
epicfaace: Why are you `gitignore`'ing pubspec.lock?

creativecreatorormaybenot: `pubspec.lock` is not needed to build a Flutter app (it will be generated when running `flutter pub get`, which **needs** to be done anyway). If we do not add the file to `.gitignore`, every PR and commit will have `pubspec.lock` changes based on the environment of the committer.

I can remove it again if you say that this intended.

epicfaace: I think we should include it. See https://dart.dev/guides/libraries/private-files#pubspeclock:

> For library packages, do not commit the pubspec.lock file.

> For application packages, do commit pubspec.lock. This file is the recommended way for non-shared resources such as applications to manage their dependencies.

creativecreatorormaybenot:
@epicfaace Done. I think that this might create some merge issues in the future, but I want to follow your advice.

xster:
checking in pubspec.lock is sensible since it locks the specific version of dependencies for everyone rather than letting everyone resolve it themselves (and potentially different versions that matches the caret semver specification)

otherwise the new constraints LG. You get collection control flows etc and there's no real reason to keep the Dart version behind the stable Flutter version.

PR #369, opened 2020-03-27
Fix missing visual feedback for `PageButton` with light background

dustincatap:
Fix missing touch feedback for `PageButton` with light background.

When `lightcolor` is `true`, it uses the `primaryColor` as the splash color, otherwise it uses `Colors.white`. Both of these colors will be applied with an alpha of `26%`.

No

 Tested on Samsung S10 (Android 29).

Old:

New:


advayDev1:
We're going to be totally revamping the UI design so please hold off on this for now.

patniemeyer:
Thank you for your effort on this but we have changed the design and so I'm going to close this now.

Issue #242, opened 2020-03-24
Small design tweaks


viviancromwell:
#174 

per this ticket, I added small tweaks in the UI that includes left and right arrows. ( this should enable both swipe gesture as well as onClick).

As the result of that, I also made the "x" close sign to be grey so the primary call to action is to continue swiping/clicking to read the content, instead of closing. Showing subtle preference what we want the users to click. 

This is not the ideal design, as too many different directions on the page.  I have gotten very little direction on the content structure from the start.  But it is improving.

A better way to do this kind of content, which I made a suggestion here. 
#240 

This is more intuitive, simple and recommended by Material Design ( arrow is the second recommendation)

viviancromwell:
@hspinks

Issue #129, opened 2020-03-21
Current design spec


hspinks:
I will keep this ticket up-to-date with the current design spec as I see it on Figma to give us something to work to as we keep the UI in line with design. I'll be adding new comments to the bottom of this ticket with each update.

hspinks:
*As of Mar 21 3pm PT:*

Onboarding flow / first-time user:
<img width="1007" alt="Screen Shot 2020-03-21 at 3 01 28 PM" src="https://user-images.githubusercontent.com/1484366/77237400-27285b00-6b85-11ea-9407-5fdfc844dbad.png">

Main screen / returning user:
<img width="490" alt="Screen Shot 2020-03-21 at 3 01 36 PM" src="https://user-images.githubusercontent.com/1484366/77237406-327b8680-6b85-11ea-8ff8-f77fc8d95fcd.png">

Protect yourself flow:
<img width="1246" alt="Screen Shot 2020-03-21 at 3 02 04 PM" src="https://user-images.githubusercontent.com/1484366/77237414-3d361b80-6b85-11ea-9e8c-ff5b65c38269.png">

Check your health flow:
<img width="1042" alt="Screen Shot 2020-03-21 at 3 01 45 PM" src="https://user-images.githubusercontent.com/1484366/77237417-4626ed00-6b85-11ea-9340-4c0dc4341994.png">

ltphan:
I will start with the main screen/returning user.

SamMousa:
So are we, or we are not doing a basic triage survey in v0.1? It was removed by @myiremark based on these specs; at the same time it was also updated by @pjayathissa (#203).

hspinks:
_As of Mar 23 9am PDT:_

Onboarding / first-time user flow:
<img width="1038" alt="Screen Shot 2020-03-23 at 9 05 32 AM" src="https://user-images.githubusercontent.com/1484366/77337222-d8de9d80-6ce5-11ea-8377-22b76c36d15f.png">
<img width="862" alt="Screen Shot 2020-03-23 at 9 05 39 AM" src="https://user-images.githubusercontent.com/1484366/77337241-e09e4200-6ce5-11ea-8069-3bdcdb7bfe0a.png">

Main screen (returning user):
<img width="433" alt="Screen Shot 2020-03-23 at 9 05 47 AM" src="https://user-images.githubusercontent.com/1484366/77337271-e98f1380-6ce5-11ea-9995-f06a7120ab62.png">

[...]

PR #373, opened 2020-03-27
Add left and right arrows to the carousel page

xster:

Resolves https://github.com/WorldHealthOrganization/app/issues/242

@viviancromwell I don't have access to your actual updated specs anywhere (https://github.com/WorldHealthOrganization/app/issues/129 doesn't seem to have it), so I'm just following your updates verbally. 

Also fixes a few nit bugs along the way:

- Passing in a BuildContext to a StatelessWidget's constructor isn't right unless you're trying some specific tree surgery type operations which isn't the case here. You generally want to access the BuildContext for the Element your widget is associated with (which is passed in during build).
- Holding state in a StatelessWidget (CarouselView) isn't right. Currently we're lucky in that no parent is causing a rebuild of the CarouselView. If there were, we'd keep building a new pageController. 

No

Added widget test. 

flutter_01

### DIFF
--- a/client/flutter/lib/carousel_page.dart
+++ b/client/flutter/lib/carousel_page.dart
@@ -7,9 +7,8 @@ import 'package:page_view_indicator/page_view_indicator.dart';
 class CarouselSlide extends StatelessWidget {
   final Widget titleWidget;
   final String message;
-  final BuildContext context;
 
-  CarouselSlide(this.context, {this.titleWidget, this.message = ""});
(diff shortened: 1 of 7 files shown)


advayDev1:
We're going to be totally revamping the UI design so please hold off on this for now.

xster:
Is the revamp spec'ed somewhere?

Miksier:
@xster 
answering your question:
#129

patniemeyer:
Thank you for your work but we have removed the carousel in the new design.  Closing this.

PR #478, opened 2020-03-27
First pass at question index page: data driven with html fragments.

patniemeyer:


patniemeyer:
image

patniemeyer:
This is just a rough cut of the question index with placeholder data source.

client/flutter/lib/components/question_index.dart
```diff
@@ -0,0 +1,114 @@
...
+        slivers: [
+          SliverAppBar(
+            backgroundColor: Colors.white,
+            flexibleSpace: FlexibleSpaceBar(
+                background: SafeArea(
+              child: Row(
+                  mainAxisAlignment: MainAxisAlignment.spaceBetween,
+                  crossAxisAlignment: CrossAxisAlignment.center,
+                  children: <Widget>[
+                    IconButton(
```
theswerd: Is this button necessary?

Normally it autogenerates a back button

patniemeyer: Flutter wasn't giving us one with this header arrangement... Perhaps I am missing something.

client/flutter/lib/components/question_data.dart
```diff
@@ -0,0 +1,52 @@
...
+              The most common symptoms of COVID-19 are:
+              <ul>
+              <li>🤒 fever</li>
+              <li>tiredness</li>
+              <li>cough</li>
+              </ul>
+              """),
+      QuestionItem(
+          title:
+          "What are coronaviruses, <b>what is COVID-19</b>, and how is it related to SARS? 2",
```
advayDev1: Do we need this suffix numbers?

advayDev1:
Please document new dependency https://pub.dev/packages/flutter_html

client/flutter/lib/components/question_data.dart
patniemeyer: I am using the title as a key and they had to be unique :)  I wasn't sure if we'd have another value to use as a key or if we should use the index order... It's just placeholder.

client/flutter/lib/components/question_index.dart
```diff
@@ -0,0 +1,115 @@
...
+                  left: 16, right: 16, top: 32, bottom: 32),
+              child: html(questionItem.body),
+            )
+          ],
+        ),
+      ),
+    );
+  }
+
+  // flutter_hmtl supports a subset of html: https://pub.dev/packages/flutter_html
```
epicfaace:
```suggestion
  // flutter_html supports a subset of html: https://pub.dev/packages/flutter_html
```

client/flutter/lib/components/question_data.dart
```diff
@@ -0,0 +1,52 @@
...
+
+/// This is just a *placeholder* for the dynamic content.
+class QuestionData {
+  static Future<List<QuestionItem>> questions() async {
+    return [
+      QuestionItem(
+          title:
+          "What are coronaviruses, <b>what is COVID-19</b>, and how is it related to SARS?",
+          body: """
+              <h2>What are the symptoms of COVID-19?</h2>
```
epicfaace: How would localization work for this? Would we need to localize this whole chunk of html, or is there a way to localize individual phrases within this single string?

kassim: By splitting it per section (subtitle, list heading, symptom fever, symptom cough, etc) the chances of breaking HTML and line-break structure lower - but it does increase complexity with more keys and less context.

I know some translation software gives warnings so these are maintained. It's not that clear from Lingo24's website that it does that.

Perhaps @clementmouchet and @sapte91 can advise?

client/flutter/lib/components/question_index.dart
patniemeyer: You were right!  I fixed this.

client/flutter/lib/components/question_index.dart
patniemeyer: Fixed!

client/flutter/lib/components/question_data.dart
patniemeyer: Good question but we haven't decided yet... working on it today.

patniemeyer: Attempting to tag @dleuck

clementmouchet: The present structure is suitable and preserves context, allowing translators to work on the complete unit. The CAT tool will take care of filtering the HTML, so there's no risk of breaking it, and segmentation will kick in for each bullet point (and sentences inside if any) so that's not an issue at all. This component doesn't have access to the context as far as I can tell on master, so I can't externalise the strings at the moment.
I'm free to work on that with you today @patniemeyer if you can.

PR #498, opened 2020-03-28
Minor reorg of files.

patniemeyer:
Minor reorg of files.


client/flutter/lib/main.dart
```diff
@@ -1,7 +1,7 @@
 import 'package:flutter/material.dart';
 import 'package:flutter/services.dart';
 import 'package:intl/intl.dart';
-import './home_page.dart';
+import 'pages/home_page.dart';
```
advayDev1: so when its a subdir, you don't use './pages'?  I do not know the convention just asking

patniemeyer: These were all automatically generated by the IDE, so I am just sticking with its recommendation :) I assume it is following the flutter format convention.

PR #560, opened 2020-03-29
Initial template for questions page dynamic content.  Added yaml pars…

patniemeyer:
Small PR to add the questions page dynamic content yaml template and yaml parser.

### DIFF
--- a/client/flutter/lib/l10n/your_questions_answered.yaml
+++ b/client/flutter/lib/l10n/your_questions_answered.yaml
@@ -1,11 +1,12 @@
 schema_version: 1
+content_version: 1
 contents:
   type: qa
   items:
     -
-    title: |
+    title_html: |
       What are coronaviruses, what is COVID-19 and how is it related to SARS?
-    body: |
+    body_html: |
       <b>What are coronaviruses, COVID-19 and how are they related to SARS?<b>
 
       <b>What is a coronavirus?<b>
@@ -27,9 +28,9 @@ contents:
       infectious than COVID-19. There have been no outbreaks of SARS anywhere in the world
       since 2003.</p>
     -
-    title: |
+    title_html: |
       Question two title...
-    body: |
+    body_html: |
       Question two body...
 
 
--- a/client/flutter/lib/l10n/your_questions_answered.yaml
+++ b/client/flutter/lib/l10n/your_questions_answered.yaml
@@ -1,0 +1,35 @@
+schema_version: 1
+contents:
+  type: qa
+  items:
+    -
+    title: |
+      What are coronaviruses, what is COVID-19 and how is it related to SARS?
+    body: |
+      <b>What are coronaviruses, COVID-19 and how are they related to SARS?<b>
+
+      <b>What is a coronavirus?<b>
+      <p>Coronaviruses are a large family of viruses which may cause illness in animals or humans.
+      In humans, several coronaviruses are known to cause respiratory infections ranging from the
+      common cold to more severe diseases such as Middle East Respiratory Syndrome (MERS) and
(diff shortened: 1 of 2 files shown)


patniemeyer:
Maybe this should not go in the i10n folder - it seems to trigger the plugin to rebuild the localized files every time you touch it, which wasn't the intent.  Probably should move to assets...

client/flutter/lib/l10n/your_questions_answered.yaml
```diff
@@ -0,0 +1,35 @@
+schema_version: 1
+contents:
+  type: qa
```
hspinks: Not a blocker, but consider making this more descriptive/generic, e.g. `accordionlist` or `questionanswer`. I'm just not sure someone looking at this down the line would interpret `qa` as question-answer. Since we're not using `type` anywhere yet anyway, it's totally up to you whether you want to change here and what you'd like to use.

```suggestion
  type: accordionlist
```

client/flutter/lib/l10n/your_questions_answered.yaml
```diff
@@ -0,0 +1,35 @@
+schema_version: 1
```
hspinks: another non-blocker, but consider using minor version?

```suggestion
schema_version: 1.0
```

advayDev1: IMO think these should be SOLELY breaking versions, and integers.  Everything else should be optional fields, like in protobufs.

client/flutter/lib/l10n/your_questions_answered.yaml
```diff
@@ -0,0 +1,35 @@
...
+contents:
+  type: qa
+  items:
+    -
+    title: |
+      What are coronaviruses, what is COVID-19 and how is it related to SARS?
+    body: |
```
advayDev1: For security every html field that will not be escaped should be suffixed _html

client/flutter/lib/l10n/your_questions_answered.yaml
```diff
@@ -0,0 +1,35 @@
+schema_version: 1
+contents:
+  type: qa
+  items:
+    -
+    title: |
```
advayDev1: Is this html or text?

client/flutter/lib/l10n/your_questions_answered.yaml
hspinks: in that case, I do propose that we add a `content_version` for eventually when we want to use caching. but that's definitely not for v1.0 and definitely not a blocker on this PR

advayDev1: Yes content_version should def be there

client/flutter/lib/l10n/your_questions_answered.yaml
patniemeyer: I was thinking of this as identifying the type of content and not the desired presentation...  I think we will only have a few of these (maybe two) and the content itself could be presented in different ways...

client/flutter/lib/l10n/your_questions_answered.yaml
patniemeyer: I will add content_version

client/flutter/lib/l10n/your_questions_answered.yaml
patniemeyer: both the title and body are html.  I will change the keys to title_html and body_html.   Sound right?

client/flutter/lib/l10n/your_questions_answered.yaml
hspinks: makes sense re type of content, I think the comment was more that `qa` might not be immediately obvious to folks that it's question-answer. not a blocker on this PR

PR #568, opened 2020-03-29
Refactor question index page for use on both data driven pages and ad…

patniemeyer:
I made the question index page more generic so that it can be re-used with both the "your questions answered" and "WHO Myth-busters" pages.  Both pages are now in place.


client/flutter/lib/pages/question_index.dart
```diff
@@ -36,19 +51,12 @@ class _QuestionIndexPageState extends State<QuestionIndexPage> {
       child: CustomScrollView(
         slivers: [
           SliverAppBar(
+            // Hide the built-in icon for now
             iconTheme: IconThemeData(
-              color: Colors.black,
+              color: Colors.transparent,
             ),
```
britannio: `automaticallyImplyLeading: false` should do the same thing.

patniemeyer: This sounds good. We should consider for cleanup after launch.

PR #588, opened 2020-03-30
Initial pass at dynamic content loading.

patniemeyer:
This adds dynamic content loading of localized YAML files with fallback to local resources.


client/flutter/lib/api/question_data.dart
```diff
@@ -0,0 +1,31 @@
+import 'package:WHOFlutter/api/content_loading.dart';
+import 'package:flutter/cupertino.dart';
+
+class QuestionData {
+  static Future<List<QuestionItem>> yourQuestionsAnswered(
+      BuildContext context) async {
+    var bundle =
```
hspinks: does this need to be wrapped in a `try`/`catch` to handle the exception thrown if there's no content bundle? or is that exception handled elsewhere?

client/flutter/lib/main.dart
```diff
@@ -43,7 +43,8 @@ class _MyAppState extends State<MyApp> {
       home: Directionality(
           child: HomePage(),
           textDirection:
-            GlobalWidgetsLocalizations(Locale(Intl.getCurrentLocale())).textDirection),
+              GlobalWidgetsLocalizations(Locale(Intl.getCurrentLocale()))
```
hspinks: are these changes coming from different editors' style preferences?

client/flutter/lib/api/content_bundle.dart
```diff
@@ -0,0 +1,38 @@
...
+      return yaml['contents']['items'];
+    } catch (err) {
+      return YamlList();
+    }
+  }
+
+  // Return the value for the specified key or the default value if not found.
+  // Note: A missing key should not prevent the bulk of the page from loading.
+  // ignore: unused_element
+  String _getString(String key, {String orDefault = "..."}) {
```
advayDev1: Maybe just delete this function since its private.

client/flutter/lib/api/question_data.dart
advayDev1: yes agree @patniemeyer

patniemeyer: 👍

client/flutter/lib/api/content_loading.dart
```diff
@@ -0,0 +1,89 @@
...
+import 'package:flutter/material.dart';
+import 'content_bundle.dart';
+
+class ContentLoading {
+  static final networkLoadingEnabled = true;
+  static final ContentLoading _singleton = ContentLoading._internal();
+
+  // TODO: Real URL
+  static final String baseContentURL =
+      'http://who.int/who/content/v1'; // no trailing
```
advayDev1:
```suggestion
      'https://who.int/covid19/content/v1'; // no trailing
```

client/flutter/lib/api/question_data.dart
```diff
@@ -0,0 +1,31 @@
...
+        await ContentLoading().load(context, "your_questions_answered");
+    try {
+      return bundle.contents
+          .map((item) =>
+              QuestionItem(title: item['title_html'], body: item['body_html']))
+          .toList();
```
daohoangson: Blindly trust network contents is dangerous. A few edge cases to consider:

- One of the `item` is not a map
- `item` doesn't have a `title_html` or `body_html` or both (!)

I think we should do some basic checks to avoid one bad item ruining the entire content bundle.

```suggestion
      return bundle.contents
          .where((item) =>
              item is Map &&
              item.containsKey('title_html') &&
              item.containsKey('body_html'))
          .map((map) => QuestionItem(
                title: map['title_html'].toString(),
                body: map['body_html'].toString(),
              ))
          .toList();
```

PR #595, opened 2020-03-30
Add endpoints for notification token and location

crazybob:
- [x] REQUIRED: Do you have an Issue **assigned to you** within the v1 milestone for this PR?  Put the Issue number here: 94

Add endpoints to capture notification token and location.

I checked in the code for s2 geometry directly since it's not in Maven. We shouldn't depend on 3rd party deployments for security reasons.

With curl. I'll follow up with an automated test.


api/who/who.proto
```diff
@@ -4,11 +4,23 @@ package who;
...
+
+  rpc putDeviceToken(PutDeviceTokenRequest) returns (Void);
+  rpc putLocation(PutLocationRequest) returns (Void);
+}
+
+message PutDeviceTokenRequest {
+  string token = 1;
+}
+
+message PutLocationRequest {
```
advayDev1: We're adding random jitter to this on the client-side right?  Maybe a comment right here to avoid any confusion.

server/appengine/src/main/java/who/Client.java
```diff
@@ -0,0 +1,46 @@
...
+import com.googlecode.objectify.condition.IfNotNull;
+
+import static com.googlecode.objectify.ObjectifyService.ofy;
+
+/**
+ * Represents an app client.
+ */
+@Entity @Cache public class Client {
+
+  @Id String uuid;
```
advayDev1: Can we consistently call this clientId here and in the header, or change the header to `Who-Client-UUID`?

server/appengine/src/main/java/who/ClientInterceptor.java
```diff
@@ -0,0 +1,50 @@
...
+      case "WEB":
+        platform = Platform.WEB;
+        break;
+      default:
+        throw new ClientException("Missing " + PLATFORM + " header.");
+    }
+
+    Client client = Client.getOrCreate(clientId, platform);
+    Client.currentClient.set(client);
+    logger.info(CLIENT_ID + ": " + client.uuid);
```
advayDev1: Can't remember if/how this appears in AppEngine logs: Would this be associated with the IP address of the http request?

api/who/who.proto
crazybob: Correct, if we're adding noise to the location, it's best to do so in the client.

server/appengine/src/main/java/who/Client.java
crazybob: This is private and in the context of the Client class, so clientId would be redundant. Best to keep "ID" in the header in case we want to use a different type of ID down the road.

server/appengine/src/main/java/who/ClientInterceptor.java
crazybob: Good point. Removed.

THIRD_PARTY_LICENSE
```diff
@@ -15,7 +15,23 @@ This repository may contain code or content licensed as follows:
    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
    See the License for the specific language governing permissions and
    limitations under the License.
-   
+
+====
+
+   Copyright 2020 Google, Inc. (S2 Geometry)
```
advayDev1: Since we aren't using a third_party directory for S2, please add a .gitattributes file in the root with one line.  
```
server/appengine/src/main/java/com/google/** linguist-vendored
```
https://github.com/github/linguist#vendored-code

PR #632, opened 2020-03-30
Add dynamic content schema version check with prompt to upgrade the app.

patniemeyer:
This adds the schema version check to the dynamic content loading with a dialog to ask the user to upgrade the app if the network content newer than supported (still falling back to the local content).  The question wrapper now also checks the content type.

image


client/flutter/lib/api/content_loading.dart
```diff
@@ -35,6 +36,18 @@ class ContentLoading {
...
+        if (err is ContentBundleVersionException) {
+          // Defer showing the dialog briefly until after screen build.
+          Future.delayed(const Duration(seconds: 1), () {
+            // TODO: Localize
+            Dialogs.showAppDialog(
+                context: context,
+                title: "App Update Required",
+                // TODO: Provide the sharing link here?
+                bodyText:
+                    "Please update to the latest version of the app in order store to receive the latest information and updates.");
```
advayDev1:
```suggestion
                    "Please update to the latest version of the app in order to receive the latest information and updates.");
```

client/flutter/lib/api/content_loading.dart
```diff
@@ -35,6 +36,18 @@ class ContentLoading {
         return await _loadFromNetwork(name, languageAndCountry);
       } catch (err) {
         print("Network bundle for $languageAndCountry not found: $err");
+        if (err is ContentBundleVersionException) {
+          // Defer showing the dialog briefly until after screen build.
+          Future.delayed(const Duration(seconds: 1), () {
+            // TODO: Localize
+            Dialogs.showAppDialog(
```
advayDev1: Should logic like this go in the widgets instead of in `lib/api`?
You could catch the exception there right?  And avoid the 1 second delay?  Seems fragile this way.

patniemeyer: You are correct to be skeptical - here are some issues: 1) It's not really the timing here as just deferring the show safely to the UI thread - a delay of zero would be equivalent.   2) We don't want to throw an exception here because we want to proceed and load the local content and just "warn" about it.  We can easily refactor this to an API that would support passing that information back, this was just the quickest / smallest change I could make this morning. 3) I don't like having the BuildContext in this class at all but it is necessary for access to the Locale information (as far as I understand) and therefore we are already sort of bound to the UI layer.  

The combo of the above lead me to thinking this was the best approach for the moment.  I will look for a better way to refactor this today if time permits.

client/flutter/lib/api/content_loading.dart
patniemeyer: fixed.

client/flutter/lib/api/content_loading.dart
```diff
@@ -35,6 +36,18 @@ class ContentLoading {
         return await _loadFromNetwork(name, languageAndCountry);
       } catch (err) {
         print("Network bundle for $languageAndCountry not found: $err");
+        if (err is ContentBundleVersionException) {
```
britannio: `on ContentBundleVersionException catch (err)` would be better

client/flutter/lib/api/content_loading.dart
patniemeyer: I added a TODO for the larger refactoring.

client/flutter/lib/api/content_loading.dart
patniemeyer: It does need to handle the exception and print the message for both... does that fall through?  (Not familiar enough with Dart exception handling :) )

britannio: Not sure what you're after so here's a DartPad example to play with https://dartpad.dev/a413da20d821e80d10f78540cec452fe

patniemeyer:
@clementmouchet heads up that this PR has some additional strings not represented in the figma design (It's a dialog pop-up).   Sorry that this process is messy.

client/flutter/lib/api/content_loading.dart
advayDev1: Pat is right let's keep it this way

advayDev1:
The build is broken.  But the preview is working.  Merging anyway

PR #664, opened 2020-03-30
Add remaining PM and Design from Index Alpha Roster

advayDev1:


content/credits.yaml
```diff
@@ -37,6 +37,19 @@ team:
...
+  - Aishwarya Vardhana
+  - Alex Hays
+  - David Kaneda
+  - Diana Alayon
+  - Graeme Fordyce
+  - Kellie Menendez
+  - Kristina Plummer
+  - Spencer Gabor
+  - Chris Hosmer
+  - Vivian Cromwell
```
hspinks:
```suggestion
  - Vivian Cromwell
  - Shilpa Apte
```

advayDev1: Sorry I will have to do it in a different PR

PR #652, opened 2020-03-30
Question page styling

theswerd:
**Please follow this checklist. Please check each appropriate box (put an 'x' or check it after creating the PR).**
- [ ] REQUIRED: Do you have an Issue **assigned to you** within the v1 milestone for this PR?  Put the Issue number here:
- [x] Provided detailed pull request description and a succinct title (consider template below for guidance).
- [x] Tested your changes, especially after any code review iterations.
- [x] Included any relevant screenshots of UI updates.
- [x] Followed the [Contributor Guidelines](https://github.com/WorldHealthOrganization/app/blob/master/docs/CONTRIBUTING.md).
- [x] Verified all contributions are properly licensed pursuant to the [LICENSE](https://github.com/WorldHealthOrganization/app/blob/master/LICENSE) file in the root of the repository.
- [x] Verified your name is in the [content/credits.yaml](https://github.com/WorldHealthOrganization/app/blob/master/content/credits.yaml) file (if you want it to be).

After all boxes above are checked, request and receive an Approved review from any team member knowledgable in the area (TODO team member list).  Once approved, the team member will assign your review to a Committer or use the `needs-merge` label.

## What does this PR accomplish?

Styled question pages
- added white background color
- removed padding between expansion panels and dividers

## Did you add any dependencies?

Nope

## How did you test the change?

iOS simulator

Simulator Screen Shot - iPhone 8 Plus - 2020-03-30 at 13 00 31
Simulator Screen Shot - iPhone 8 Plus - 2020-03-30 at 13 00 33
Simulator Screen Shot - iPhone 8 Plus - 2020-03-30 at 13 00 36

PR #684, opened 2020-03-31
CI - Run git diff to make sure Podfiles, etc. have been updated on a new dependency add

### LINKED ISSUE #647: CI should check that files such as Podfile, Podfile.lock are properly updated
When we initially integrated analytics through #512, we added a Flutter dependency `firebase_analytics` -- but we did not run the iOS build; running the iOS build would have updated the Podfile / Podfile.lock, and these updates did not make it into that PR. I fixed this in #621 -- but it would be good if the CI catches this in the future.

The CI build script should:
- Run flutter build (iOS / Android)
- Do a `git diff` and make sure no files have changed locally
- If files have changed locally, fail the CI process and print out what has changed.

epicfaace:
fixes #647

Runs git diff on CI to make sure additional files such as Podfile, Podfile.lock, pubspec.lock have been updated when a new dependency has been added to flutter.

This will ensure that people are actually committing changes to files such as Podfile.lock -- which isn't caught by the current CI process.

I've also updated Podfile.lock and pubspec.lock which weren't updated properly earlier, causing the initial check added by this PR to fail.


epicfaace:
[The failed build](https://github.com/WorldHealthOrganization/app/pull/684/checks?check_run_id=547324886) shows why this change is important -- apparently #652 caused Podfile.lock to revert back to using CocoaPods 1.8.4 from 1.9.1.

.github/workflows/flutter-build.yml
```diff
@@ -41,3 +42,4 @@ jobs:
         working-directory: ./client/flutter
       - run: flutter build appbundle
         working-directory: ./client/flutter
+      - run: git diff --exit-code
```
hspinks: do we need this for Android if this is only meant to catch Podfile updates? probably not a problem to have it here / not a blocker, just checking

epicfaace: This catches updates both to `Podfile` and also for `pubspec.lock` (which will matter for Android builds). That being said, I'm not sure if there are any Android-specific files that flutter updates.

PR #708, opened 2020-03-31
No server content, ToS/PP url, COVID-19 term, disable onboarding

advayDev1:


LICENSE
```diff
@@ -1,4 +1,4 @@
-Built by the WHO COVID App Collective (see content/credits.yaml).
+Built by the WHO COVID-19 App Collective (see content/credits.yaml).
```
hspinks: does the switch to COVID-19 apply here too?

client/flutter/lib/pages/onboarding/legal_landing_page.dart
```diff
@@ -38,12 +38,20 @@ class LegalLandingPage extends StatelessWidget {
...
                   TextSpan(
-                    text: "By proceeding, you agree with our "
+                    text: "By proceeding, you agree to our "
                   ),
                   LinkTextSpan(
-                    text: "terms of service",
+                    text: "Terms of Service",
                     style: TextStyle(decoration: TextDecoration.underline),
-                    url: "https://who.int",//TODO: REPLACE WITH TERMS OF SERVICE LINK
+                    url: "https://whocoronavirus.org/terms",
```
hspinks: I wasn't sure if @crazybob said this or https://whocoronavirus.org/tos on the phone just now... I'm sure it can be set up either way

advayDev1: it doesn't matter lol

LICENSE
advayDev1: our credits screen says COVID-19 - should use one name

PR #722, opened 2020-03-31
Simply achieved click debounce

Moosphan:
- [x] REQUIRED: Do you have an Issue **assigned to you** within the v1 milestone for this PR?  Put the Issue number here:

See [#714](https://github.com/WorldHealthOrganization/app/issues/714)
Provide a simple way to achieve click debounce. The problem you can see in the following screenshot:

Kapture 2020-03-31 at 21 16 24

After debounce, you can see the effect in the following screenshot on my phone:

Kapture 2020-03-31 at 21 09 14

However, this way i used may not good enough. So i just apply to the share button in home page. We know that it has many similar buttons need to achieve debounce. If you have better way, let me know and i will change it.

No.

By running on my phones.


client/flutter/lib/utils/debounce_handler.dart
```diff
@@ -0,0 +1,19 @@
+
+import 'dart:async';
+import 'package:flutter/material.dart';
+
+/// Handle debounce when we click multiple times quickly.
+class DebounceHandler {
```
advayDev1: Is this the best practice generally, or your own solution?

dna-f: Hi guys! Why don't we use streams and the debounce extension of rxdart?

patniemeyer: I am a huge fan of RxDart and as we add complexity to the app I would like to see us introduce it where appropriate.  However if we need button debouncing in these simple cases it might just be best to do something ad-hoc for now.  Rather than a timer I would probably just add a `fooEnabled` state variable to the containing class to govern the button state.  This is about as simple as it gets and can be controlled by the logic as necessary.  If we get into more complex situations composing data sources and publishing state to the entire app we should definitely use streams and probably RxDart.

Moosphan:
I'm going to close this PR and thanks for @patniemeyer 's suggestion😄

PR #729, opened 2020-03-31
[feat] Server endpoint skeleton for global COVID-19 stats

advayDev1:
Defines a new endpoint for global COVID-19 stats.  Our use of a proxied backend, a static file we push once a day, etc. can be controlled behind this facade.  We will use a single API to get both global stats and local stats given location.

Test:
````
curl -i     -H 'Content-Type: application/json' -H 'Who-Client-ID: 00000000-0000-0000-0000-000000000000' -H 'Who-Platform: IOS' -X POST -d ''     http://localhost:8080/WhoService/getCaseStats

HTTP/1.1 200 OK
Content-Type: application/json
Server: Development/1.0
Date: Tue, 31 Mar 2020 20:22:24 GMT
Cache-Control: no-cache
Expires: Mon, 01 Jan 1990 00:00:00 GMT
Content-Length: 123

{"globalStats":{"lastUpdated":1585641600000,"cases":719758,"deaths":33673,"recoveries":-1,"attribution":"WHO"},"ttl":21600}
````

### CODE REVIEW

File: server/appengine/src/main/java/who/PublicWhoServiceImpl.java [hunk 1]
```diff
@@ -0,0 +1,26 @@
+package who;
+
+// All responses are Client-independent.
+// No Client instance is available in this service.
+public class PublicWhoServiceImpl implements PublicWhoService {
+
+  @Override public GetGlobalStatsResponse getGlobalStats(Void request) {
+    // TODO: Configure cache response headers. RFC 2616 Section 9.5 says we can cache
+    // POST requests.  Or we can enable GET method in present.rpc.RpcFilter.
```
epicfaace: I think it makes more sense (from an API perspective) to make this a GET method.

advayDev1: It does, but RpcFilter forced POST right now.

api/who/who.proto
```diff
@@ -4,12 +4,22 @@ package who;
...
 
   // The client should not pass a precise location.
   rpc putLocation(PutLocationRequest) returns (Void);
+
+  // TODO: Add location-specific stats.
+}
+
+// These responses are publicly cacheable and clients must
+// not send a client ID header.
+service PublicWhoService {
```
crazybob: I'd prefer we keep things simple and just use one service for now. All rpcs should always be idempotent. Caching policies for individual rpcs will likely vary.

api/who/who.proto
```diff
@@ -21,6 +31,21 @@ message PutLocationRequest {
...
 
+message GetGlobalStatsResponse {
+  CaseStats stats = 1;
+  string attribution = 2;
+}
+
+// Statistics in a single or aggregate jurisdiction.
+message CaseStats {
+  // msec since epoch.
+  int64 lastUpdated = 1;
```
crazybob: I usually use uint64 for timestamps.

server/appengine/src/main/java/who/PublicRpcInterceptor.java
```diff
@@ -0,0 +1,20 @@
+package who;
+
+import present.rpc.ClientException;
+import present.rpc.RpcInterceptor;
+import present.rpc.RpcInvocation;
+
+/**
+ * For publicly-cacheable RPCs.
+ */
+public class PublicRpcInterceptor implements RpcInterceptor {
```
crazybob: We shouldn't need this class. Best to isolate logic to ClientInterceptor.

server/appengine/src/main/java/who/PublicWhoRpcFilter.java
```diff
@@ -0,0 +1,26 @@
...
+import com.google.inject.Inject;
+import com.google.inject.Singleton;
+import present.rpc.LoggingInterceptor;
+import present.rpc.RpcFilter;
+import present.rpc.RpcInterceptorChain;
+
+/**
+ * Configures RPC services for publicly-cacheable APIs.
+ */
+@Singleton public class PublicWhoRpcFilter extends RpcFilter  {
```
crazybob: We don't need this class. For future reference, we need just one RpcFilter class for the whole app. You can call service() multiple times.

server/appengine/src/main/java/who/PublicWhoServiceImpl.java
```diff
@@ -0,0 +1,26 @@
...
+
+    // TODO: Get from data source.
+    return new GetGlobalStatsResponse.Builder()
```
crazybob: Present RPC is designed to be protocol independent. Perhaps we could add a ttl field to the response message?

api/who/who.proto
```diff
@@ -21,6 +31,21 @@ message PutLocationRequest {
   double longitude = 3;
 }
 
+message GetGlobalStatsResponse {
```
crazybob: Are these intended to be global stats? Guess we can add fields for local stats later?

advayDev1: Yes global only.

server/appengine/src/main/java/who/PublicWhoServiceImpl.java
advayDev1: done

api/who/who.proto
advayDev1: Per discussion down to one api for both local and global stats, thus removing the utility of a publicly-cacheable API/service.

advayDev1:
PTAL @crazybob - per our discussion, Benefits of public-caching are minor and outweighed by having a single roundtrip for both global and local (eventual) stats. Thus the single api no longer need be prevented from reading the client-id, removing the usefulness of two separate services.

api/who/who.proto
```diff
@@ -21,6 +25,30 @@ message PutLocationRequest {
   double longitude = 3;
 }
 
+message GetCaseStatsResponse {
+  CaseStats globalStats = 1;
+
+  // TODO: Add local stats fields too, potentially at multiple administrative levels.
+
+  // Clients should not re-request stats until at least ttl seconds.
+  int64 ttl = 2;
```
crazybob: I usually default to uint32 or 64 unless a number can be negative.

PR #741, opened 2020-03-31
Recreate WHO logo for legal page using localized text and mark image.

patniemeyer:
This replaces the static WHO.jpg on the legal page with localized text and the graphic-only `mark` asset.  I'm not certain of the best approach for this that would still allow the text to wrap so I have implemented it this way as a first pass.  The text may still break inelegantly in other languages.  Normally we'd have a localized image asset for this I guess...

This is what it looks like on large and small screens after the conversion:

image

### DIFF
--- a/client/flutter/lib/constants.dart
+++ b/client/flutter/lib/constants.dart
@@ -14,3 +14,7 @@ double contentScale(BuildContext context) {
   const short = 480.0;
   return ((height - short) / (tall - short)).clamp(0.0, 1.0);
 }
+
+double contentScaleFrom(BuildContext context, double low, double high) {
+  return low + contentScale(context) * (high-low);
+}
--- a/client/flutter/lib/pages/onboarding/legal_landing_page.dart
+++ b/client/flutter/lib/pages/onboarding/legal_landing_page.dart
@@ -19,14 +19,46 @@ class LegalLandingPage extends StatelessWidget {
         child: Column(
           mainAxisAlignment: MainAxisAlignment.center,
           children: <Widget>[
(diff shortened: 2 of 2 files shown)


patniemeyer:
@clementmouchet Or is it possible that we do not want to localize this asset and we are ok with it being in English?  If that is case we still need the asset with a transparent background since it doesn't match the screen background.

crazybob:
We should not render the text ourselves, logos should be pixel perfect. I don’t see anything about translating the logo in their guidelines—normally orgs don’t. There should be an svg with a transparent background in the identity folder.

crazybob:
It looks like they do have translated logo types. We shouldn't try to recreate it ourselves—it doesn't look right, and this will come up in the design review. Let's stick to English for now and plan to incorporate their SVGs for other languages.

dleuck:
It should be a localized image resource. If they don't have variants for all locales I can localize it and set the type correctly for other locales.

dleuck:
Ideally it would be a localized SVG.

crazybob:
[WHO-guidelines-EN4-LOGO.PDF](https://github.com/WorldHealthOrganization/app/files/4412030/WHO-guidelines-EN4-LOGO.PDF)

dleuck:
> [WHO-guidelines-EN4-LOGO.PDF](https://github.com/WorldHealthOrganization/app/files/4412030/WHO-guidelines-EN4-LOGO.PDF)

Perfect! Are these localized versions available as SVGs? If not I can extract them from the PDF.

patniemeyer:
I'll close this PR and assign the issue to @dleuck

Issue #730, opened 2020-03-31
Deployment: update repo with changes from WHO


epicfaace:
WHO has made several changes (some through XCode, some involving Android release configs) since they have deployed. We need to update the repo (or the release branch) with these changes so that we have these changes in source control.

Also, make a "cut" script that swaps in dev with prod config.

hspinks:
@epicfaace thanks for noting this - can you document those changes either here or in Slack?

epicfaace:
I was sent a patch file -- I will apply them to a PR and link it to this

brunobowden:
@epicfaace - please work with @patniemeyer to get this merged in. @patniemeyer will do the build and upload to App Store.

hspinks:
Removing this from the `v1.0` milestone as our current process seems to be sufficient to get this to the app stores (see https://github.com/WorldHealthOrganization/app/pull/745#issuecomment-609111615)

advayDev1:
We now have a full build in the private repo, which from here on it we will use the app-build repo.

PR #745, opened 2020-04-01
Add scripts to swap in prod firebase config before cutting a release branch

epicfaace:

deploy-prod.sh can be run right before cutting a release branch so that the right
firebase config can be swapped in. We can also modify this to make other changes
that may need to be modified from dev -> prod.

for #730

### DIFF
--- a/config/GoogleService-Info.plist
+++ b/config/GoogleService-Info.plist
@@ -1,0 +1,36 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+	<key>CLIENT_ID</key>
+	<string>936622942451-bhjm61j553i0soe10b7vou3ebratp6bn.apps.googleusercontent.com</string>
+	<key>REVERSED_CLIENT_ID</key>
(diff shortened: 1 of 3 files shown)


hspinks:
I really like this idea of having a script to automate these tasks to avoid any possible human error on this. Where did we shake out on the conversation of including these files in the repo in general?

epicfaace:
> I really like this idea of having a script to automate these tasks to avoid any possible human error on this. Where did we shake out on the conversation of including these files in the repo in general?

These files are already included in the repo, as part of the RC branch, so it should be fine to keep them in the repo.

It's good to keep them in the repo so that they can be version-controlled and the appropriate flags / settings are also included in version control.

hspinks:
I think that what's implemented here is exactly right for a pre-deploy process/hook, but I think this should be held until the multi-repo flow proposed in #854 is implemented, since this script & the prod config files should be in the `app-private` repo. I'm moving this out of the `v1.0` milestone since this isn't necessary for launch - our existing manual process can work fine if this isn't merged by then.

> These files are already included in the repo, as part of the RC branch, so it should be fine to keep them in the repo.

These files aren't sensitive, which is why it's ok to be in the RC branch. But it's still a little bit better to make someone dig for them than to put them in `master` where every single person forking the repo has those files in their working copy. This won't stop someone who really wants to find those config files, but this isn't a huge security concern. Since the prod build will happen in the `app-private` repo anyway to make sure the proper assets are included, it makes sense for this pre-deploy script to be there, too.

hspinks:
I'm going to close this, since the prod firebase config can be included in the overlay repo as proposed in #854

PR #747, opened 2020-04-01
[feat] Latest Numbers page

### LINKED ISSUE #521: Latest Numbers Tile
For v1: A full page web view of an external URL

advayDev1:
Closes #521 

Someone else will have to make it look nice.

Stats from our server.

TODO: Caching

Screen Shot 2020-03-31 at 7 31 34 PM

### DIFF
--- a/client/flutter/lib/api/who_service.dart
+++ b/client/flutter/lib/api/who_service.dart
@@ -33,6 +33,18 @@ class WhoService {
     return true;
   }
 
+
+  static Future<Map<String,dynamic>> getCaseStats() async {
+    Map<String, String> headers = await _getHeaders();
+    var url = '$serviceUrl/getCaseStats';
+    var response = await http.post(url, headers: headers, body: '');
+    if (response.statusCode != 200) {
+      throw Exception("Error status code: ${response.statusCode}");
+    }
+    // TODO: Should use protobuf.
+    return jsonDecode(response.body);
+  }
+
   static Future<Map<String, String>> _getHeaders() async {
     var clientId = await UserPreferences().getClientUuid();
     var headers = {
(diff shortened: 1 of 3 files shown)


client/flutter/lib/api/who_service.dart
```diff
@@ -33,6 +33,18 @@ class WhoService {
     return true;
   }
 
+
+  static Future<Map<String,dynamic>> getCaseStats() async {
+    Map<String, String> headers = await _getHeaders();
+    var url = '$serviceUrl/getCaseStats';
+    var response = await http.post(url, headers: headers, body: '');
```
hspinks, line 40: why did we set up this API endpoint as a `POST` instead of a `GET` request?

client/flutter/lib/pages/latest_numbers.dart
```diff
@@ -0,0 +1,144 @@
...
+    final S localized = S.of(context);
+
+    return PageScaffold(context,
+        title: "Latest Numbers",
+        showShareBottomBar: false,
+        body: [
+          FutureBuilder(
+              future: WhoService.getCaseStats(),
+              builder: (context, snapshot) {
+                final ts = snapshot.hasData ? new DateTime.fromMillisecondsSinceEpoch(snapshot.data['globalStats']['lastUpdated']) : DateTime.now();
```
hspinks, line 42: out of scope on this PR, but probably a good idea longer term to do some validation that the data exists in this structure across all of the 2nd level keys (e.g. ensuring `snapshot.data['globalStats']` is not null before trying to get a key on it)

advayDev1: Will fix now.

client/flutter/lib/api/who_service.dart
advayDev1: Its how the present RPC framework is.  Only POST is allowed.  It was discussed in the server PR.

PR #777, opened 2020-04-01
[feat] Transmit opt-in location, as jittered [fix] COVID-19 app name [

advayDev1:
- Need to target ANdroid 10 to avoid spurious background location permission
- Change the COVID-19 App -> COVID-19
- Allow you to accept location for content even if not just for notifications
- Don't let location failures break onboarding


client/flutter/lib/pages/onboarding/notifications_page.dart
```diff
@@ -29,7 +29,7 @@ class _NotificationsPageState extends State<NotificationsPage> {
   }
 
   void _skipNotifications() async {
-    Navigator.pop(context);
+    _complete();
```
hspinks, line 32: is this tested?

PR #797, opened 2020-04-01
Fix: app not showing the home page after allowing onboard location

marctan:
- [ ] REQUIRED: Do you have an Issue **assigned to you** within the v1 milestone for this PR?  Put the Issue number here:

Only reproducible if Location Services is Off. 
This will fix a bug where the users is not forwarded to the home page after clicking the "Allow Location Sharing" button in onboarding page. This will also fix the crash when clicking "No Thanks" in the location service request popup.

No

Tested with Location Services turned off using devices below:
Pixel - Android 7.0
Redmi Note2 - Android 5.1

No Response after clicking ok issue:
no response

Crash issue:
crash

FIXED:
fix

### DIFF
--- a/client/flutter/lib/pages/onboarding/location_sharing_page.dart
+++ b/client/flutter/lib/pages/onboarding/location_sharing_page.dart
@@ -33,12 +33,16 @@ class _LocationSharingPageState extends State<LocationSharingPage> {
     try {
       await Location().requestPermission();
       if (await Location().hasPermission() == PermissionStatus.granted) {
-        LocationData location = await Location().getLocation();
-        Map jitteredLocationData = JitterLocation().jitter(location.latitude, location.longitude, 5/*kms refers to kilometers*/);
-        
-        await WhoService.putLocation(
-          latitude: jitteredLocationData['lat'],
-          longitude: jitteredLocationData['lng']);
+        if (await Location().requestService()) {
+          LocationData location = await Location().getLocation();
+          Map jitteredLocationData = JitterLocation().jitter(
+              location.latitude, location.longitude,
+              5 /*kms refers to kilometers*/);
+
+          await WhoService.putLocation(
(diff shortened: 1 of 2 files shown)


client/flutter/lib/pages/onboarding/location_sharing_page.dart
```diff
@@ -33,12 +33,16 @@ class _LocationSharingPageState extends State<LocationSharingPage> {
...
+              location.latitude, location.longitude,
+              5 /*kms refers to kilometers*/);
+
+          await WhoService.putLocation(
+              latitude: jitteredLocationData['lat'],
+              longitude: jitteredLocationData['lng']);
+        }
       }
     } catch(_) {
       // ignore for now.
```
advayDev1, line 48: Why is the finally { _complete(); } not enough?

marctan: > Why is the finally { _complete(); } not enough?

If Location Service is off and the "await Location().getLocation()" is directly called, it will never return thus it will not reach the subsequent code (e.g finally clause).  We should call Location().requestService() first to fix this.

advayDev1: Got it!  That def seems like a bug in getLocation to never terminate, but your fix looks good.

britannio:
I can reproduce that crash, not sure what the culprit is though.

client/flutter/lib/pages/onboarding/location_sharing_page.dart
britannio: https://github.com/WorldHealthOrganization/app/blob/master/client/flutter/lib/pages/onboarding/location_sharing_page.dart#L36
The future on this line isn't finishing. 
I tested the change and it all works now, no crash and after location is retrieved the page is closed.

PR #815, opened 2020-04-01
Turn on content loading / remove incorrect content

### LINKED ISSUE #813: Turn on server side access

advayDev1:
For now we'll just get 404s.  But it means that Server side it doesn't need to be done by 5pm.

Closes #813


client/flutter/lib/api/content_loading.dart
```diff
@@ -6,12 +6,12 @@ import 'package:flutter/material.dart';
 import 'content_bundle.dart';
 
 class ContentLoading {
-  static final networkLoadingEnabled = false;
+  static final networkLoadingEnabled = true;
   static final ContentLoading _singleton = ContentLoading._internal();
 
   // TODO: Real URL
```
PieterAelse, line 12: Can this `TODO` now be removed? 🙂

advayDev1: No, won't remove until confirmed with Bob

PR #826, opened 2020-04-02
P0 fix: add MediaQuery.textScaleFactorOf to page button text scale factor

britannio:
- [x] REQUIRED: Do you have an Issue **assigned to you** within the v1 milestone for this PR?  Put the Issue number here: https://github.com/WorldHealthOrganization/app/issues/822

- Closes https://github.com/WorldHealthOrganization/app/issues/822
- Uses `MediaQuery.textScaleFactorOf(context)` to incorporate the device's font scale where `textScaleFactor` was specified.

No

Viewing the home page and the travel advice page on my phone with different system font sizes before and after the fix to verify that the text now scales according to the system font size. This was conducted on my Android device.

Affected pages:
Screenshot_20200402-124315
Screenshot_20200402-124318
Screenshot_20200402-124304
Screenshot_20200402-124307


client/flutter/lib/components/page_button.dart
```diff
@@ -45,7 +45,8 @@ class PageButton extends StatelessWidget {
             children: <Widget>[
               Text(
                 this.title,
-                textScaleFactor: 1.0 + 1.0 * contentScale(context),
+                // textScaleFactorOf is for the device font size
+                textScaleFactor: 1.0 + contentScale(context) * MediaQuery.textScaleFactorOf(context),
```
hspinks, line 49: should we consider adding `MediaQuery.textScaleFactorOf(content)` directly into our `contentScale()` method? that way it would apply everywhere we use contentScale, which I think is most of the places we're editing `textScaleFactor`. we'd definitely need to review everything across the app after making that change, but something to consider

britannio: As long as `contentScale()` is only used for overriding `textScaleFactor`,

britannio: never overriding `textScaleFactor` is the way to go in my opinion.

patniemeyer: Yah, we should not be setting `testScaleFactor` for this as that defeats the system adaptive text size.  We should have used a `fontSize` instead.  *However* these thing are really more like titles and and I don't know that it is expected that UI elements change text size with the adaptive text anyway.  Allowing these to change size is going to complicate @clementmouchet 's work on getting these button/section titles to fit in all languages.  

We should get some input on whether these particular UI elements would be expected to change size with the adaptive text or whether only "content" text should adjust size automatically.  If they need to be adaptive the simplest solution is to remove the `textScaleFactor` and just use a font that matches the current size (the scaling factor can be kept, just used on the font size instead).  

Longer term I think we should switch to the `auto_text_size` package and (per discussion with @clementmouchet ) perhaps include some per-language hints about size, weight, and max lines within the ARB files in the form of annotations.

dleuck: I suggest the latter. I think you will be OK for the six locales the app supports. Russian may wrap in a few places but that is OK as long as its not in a logo.

patniemeyer: Ok, after discussing with @dleuck it sounds like we *should* allow these button/section titles to update with the adaptive text.  So I think the solution for short term is to switch from the `textScaleFactor` to `fontSize` just trying to match the existing (and using our `contentScale()` on the `fontSize` just as we did on the `textScaleFactor`.

britannio: Can I create a new pr that uses `auto_text_size`? @patniemeyer

hspinks: @patniemeyer FYI: https://github.com/WorldHealthOrganization/app/issues/884#issuecomment-609462881

client/flutter/lib/components/page_button.dart
```diff
@@ -47,7 +47,8 @@ class PageButton extends StatelessWidget {
...
```
patniemeyer, line 51: Sorry for not catching this earlier but - the way the `textScaleFactorOf()` is being applied on this line it only multiplies the second term.  There should be parens around both terms as in the fix below.

britannio: My bad, you are correct

PR #847, opened 2020-04-03
Client to use Updated Endpoints

advayDev1:


theswerd:
Tested on iPhone 11 Pro simulator

PR #875, opened 2020-04-05
[feat] Use city center lat,lon instead of jittered

### LINKED ISSUE #872: Use city locations instead of jittered GPS on the client
* Prevents jitter from sending you into the wrong political area for targeting
* Abundance of caution for low pop density areas

One possibility:
```
Real GPS lat, lon
 -> platform reverse geolocation
    -> Street Address, Locality (City), SubAdmin (County), Admin (State), CountryCode
        -> Discard street address, do platform geolocation of the rest
            -> City center lat, lon
               -> Send CityLat, CityLon, Locality (City), SubAdmin (County), Admin (State), CountryCode for notification/content targeting; (discard real GPS lat, long)
```

advayDev1:
Closes #872

TESTED:
ANDROID:

w/ loc services:
allow (app) -> allow (os)
allow (app) -> deny (os)
skip (app)

w/o loc services:
allow (app) -> allow (os) -> turn on services (os)
allow (app) -> allow (os) -> keep off services (os)
allow (app) -> deny (os)
skip (app)

IOS: UNTESTED

Dependencies:
location_permissions, geolocator, google_api_availability all 99-100 score by verified dev baseflow.com


.github/workflows/flutter-build.yml
```diff
@@ -36,7 +36,7 @@ jobs:
       - uses: actions/setup-java@v1
         with:
           java-version: '12.x'
-      - run: git clone https://github.com/flutter/flutter.git --depth 1 -b v1.12.13+hotfix.5 _flutter
+      - run: git clone https://github.com/flutter/flutter.git --depth 1 -b v1.12.13+hotfix.6 _flutter
       - run: echo "::add-path::$GITHUB_WORKSPACE/_flutter/bin"
```
brunobowden, line 40: Agreed that we should change from beta (hotfix 5) to stable (hotfix 6). I'm wondering if we should go to hotfix 9 as this includes some other fixes. Though Flutter has a policy of very limited scope of hotfixes, I still have consternation as the latest hotfix is from April 1st and only a few days old: https://github.com/flutter/flutter/wiki/Hotfixes-to-the-Stable-Channel#hotfix9-april-01-2020

.github/workflows/flutter-build.yml
```diff
@@ -21,7 +21,7 @@ jobs:
...
```
brunobowden, line 25: Extract the flutter version to a common variable, such as:
https://github.com/argoproj/argo/blob/master/examples/arguments-parameters.yaml#L13

Also add comment:
```
Update flutter-test.yml file to use same Flutter version
```

client/flutter/lib/pages/onboarding/location_sharing_page.dart
```diff
@@ -35,16 +35,52 @@ class _LocationSharingPageState extends State<LocationSharingPage> {
...
+              longitude: placeCityCenter.position.longitude,
+              countryCode: myPlace.isoCountryCode,
+              adminArea: myPlace.administrativeArea,
+              subadminArea: myPlace.subAdministrativeArea,
+              locality: myPlace.locality);
         }
       }
-    } catch(_) {
+    } catch (_) {
       // ignore for now.
```
brunobowden, line 84: Can we log errors using Google Analytics? I have concerns that Apple geocoder and reverse geocoder may not be as comprehensive at Google's, particularly in developing countries. I would at least like to be able to measure problems like that so they don't silently fail. If the problem is widespread in certain areas, we might look at using Google's Geocoding API so that we have consistency across both platforms.

https://developers.google.com/maps/documentation/geocoding/intro

client/flutter/pubspec.lock
```diff
@@ -64,6 +64,13 @@ packages:
       url: "https://pub.dartlang.org"
     source: hosted
     version: "0.16.1"
+  equatable:
```
brunobowden, line 67: I can't see Equatable being used anywhere?

.github/workflows/flutter-test.yml
```diff
@@ -21,7 +21,7 @@ jobs:
       - uses: actions/setup-java@v1
         with:
           java-version: '12.x'
-      - run: git clone https://github.com/flutter/flutter.git --depth 1 -b v1.12.13+hotfix.5 _flutter
+      - run: git clone https://github.com/flutter/flutter.git --depth 1 -b v1.12.13+hotfix.6 _flutter
```
brunobowden, line 24: Doesn't seem like a good way to share parameters across yaml files, so don't worry about it

client/flutter/pubspec.lock
advayDev1: It is a transitive dep.  pubspec.lock is auto generated from pubspec.yaml

.github/workflows/flutter-build.yml
advayDev1: What I'll do is put the comment in pubspec.yaml instead.  (And not doing the variable due to your later comment)

client/flutter/lib/pages/onboarding/location_sharing_page.dart
advayDev1: Thanks Bruno I agree about possibly using Google Maps on both platforms - we’ll need to use api keys with a quota if we don’t use the platform provider, so let’s not do it tonight. I’ve made a bug #876 to log all errors in analytics (it’s a bigger problem than just location APIs, and we need to think though how to ensure no exception text PII gets sent to google analytics).

However, on the issue of geocoders’ lack of comprehensiveness, we will not see that as errors. Instead we will see it as a partially incomplete location stored on the server (lacking city, county, etc. but containing country). We cannot distinguish between genuine lack of detail (such as unincorporated areas of US counties) and the geocoder’s inability to discern such detail. We could however run a statistical test comparing iOS to Google Maps later if needed.

.github/workflows/flutter-build.yml
advayDev1: Let me run some tests first.  We had an issue with very recent stable builds breaking us before.

client/flutter/lib/pages/onboarding/location_sharing_page.dart
brunobowden: Country code is the main thing. Not sure when WHO will be ready for even country specific notifications.

.github/workflows/flutter-build.yml
advayDev1: Alright we can move up to hotfix 8 with the build working fine.

Think we should give hotfix 9 a bit of time to bake before we upgrade to it.

advayDev1:
PTAL

brunobowden:
I'm good with hotfix 8 then

brunobowden:
LGTM

PR #834, opened 2020-04-02
update info plist to fix push notification permissions

epicfaace:
## What does this PR accomplish?

Update info plist to fix app push notification permissions

Fixes error,

```
Missing Info.plist value. The Info.plist key 'BGTaskSchedulerPermittedIdentifiers' must contain a list of identifiers used to submit and handle tasks when 'UIBackgroundModes' has a value of 'processing'. For more information, refer to the Information Property List Key Reference at https://developer.apple.com/library/ios/documentation/General/Reference/InfoPlistKeyReference/Introduction/Introduction.html.
```

See https://github.com/FirebaseExtended/flutterfire/issues/1694 for the fix

PR #887, opened 2020-04-05
Update Info.plist with changes in v0.4.1 release

epicfaace:
Adding in these changes that were added to the v0.4.1 release branch so they're on master:

- Update info plist to fix push notification permissions (#834)
- Change CFBundleDisplayName to WHO COVID-19 (#835)

### DIFF
--- a/client/flutter/ios/Runner/Info.plist
+++ b/client/flutter/ios/Runner/Info.plist
@@ -19,7 +19,7 @@
 	<key>NSLocationAlwaysAndWhenInUsageDescription</key>
 	<string>Your location is used to deliver you the most relevant information and push notifications to your area.</string>
 	<key>CFBundleDisplayName</key>
-	<string>COVID-19</string>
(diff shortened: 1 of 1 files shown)


client/flutter/ios/Runner/Info.plist
```diff
@@ -19,7 +19,7 @@
 	<key>NSLocationAlwaysAndWhenInUsageDescription</key>
 	<string>Your location is used to deliver you the most relevant information and push notifications to your area.</string>
 	<key>CFBundleDisplayName</key>
-	<string>COVID-19</string>
+	<string>WHO COVID-19</string>
```
britannio, line 22: Is this being changed on Android too? And is that still short enough to not be cut off by the IOS launcher?

advayDev1: This is just a merge so we shouldn;'t change anythig here

PR #897, opened 2020-04-05
Defer FCM autoinit until ToS approval

### LINKED ISSUE #870: FCM no autoinit, defer until ToS Approval
FCM tokens should be generated only when the user asks for notifications.
See for example:
https://firebase.google.com/docs/cloud-messaging/android/client#prevent-auto-init

Please verify by looking at intercepted traffic on Android and iOS.

epicfaace:

Fixes #870 by deferring FCM autoinit until ToS approval has been reached. I followed the instructions listed in:

Android: https://firebase.google.com/docs/cloud-messaging/android/client#prevent-auto-init

iOS: https://firebase.google.com/docs/cloud-messaging/ios/client#prevent_auto_initialization

No

Haven't tested this on a device yet -- will remove the draft status once that is done. I can test this on Android, but need someone else to help test this on iOS.

testing protocol originally done by @advayDev1 to reproduce the issue:
1.  open the app on an iOS simulator and intercept its network requests
2. let the app sit on the ToS screen and never pass it
3. watch for the following logs (which shows that FCM autoinit is happening):

```
COVID-19 (iOS)
Total: 8.46 kB sent, 27.9 kB received
	Summary of 3 connections
	Outgoing to (null) (173.194.202.188), Port 5228, Protocol TCP (6), 937 B sent, 4.5 kB received
	Outgoing to mtalk.google.com (173.194.202.188), Port 5228, Protocol TCP (6), 937 B sent, 4.5 kB received
	Summary of 4 connections
	Outgoing to device-provisioning.googleapis.com (172.217.5.202), Port https (443), Protocol TCP (6), 1.3 kB sent, 5.01 kB received
	Outgoing to firebaseinstallations.googleapis.com (172.217.5.74), Port https (443), Protocol TCP (6), 1.47 kB sent, 4.80 kB received
	Outgoing to firebaselogging-pa.googleapis.com (172.217.14.74), Port https (443), Protocol TCP (6), 1.51 kB sent, 4.58 kB received
	Outgoing to fcmtoken.googleapis.com (172.217.11.74), Port https (443), Protocol TCP (6), 2.09 kB sent, 4.99 kB received
	Outgoing to (null) (172.217.5.78), Port https (443), Protocol TCP (6), 1.15 kB sent, 4.00 kB received
	Outgoing to app-measurement.com (172.217.5.78), Port https (443), Protocol TCP (6), 1.15 kB sent, 4.00 kB received
```


patniemeyer:
@epicfaace @cbgrey This looks right to me although I can't tell from the docs if we can actually disable only the FCM messaging without also disabling the analytics (the example does them together).   

More generally - We do not upload the token to our servers until the user grants the notify permission, so I'm not sure that this is really necessary / beneficial unless we also disable analytics... They both probably use the same firebase token, don't they?   i.e. If the point is to prevent firebase registering a token for the client we may only have one lever for these two paths.  If the point is just to prevent the WHO from sending the client messages there is no chance of that without their consent on two levels: first the notif permission and second the uploading of the token to our servers.

client/flutter/lib/pages/onboarding/notifications_page.dart
```diff
@@ -40,6 +40,7 @@ class _NotificationsPageState extends State<NotificationsPage> {
     }
     
     await UserPreferences().setNotificationsEnabled(true);
+    await _firebaseMessaging.setAutoInitEnabled(true);
```
epicfaace, line 43: Move this call to the point when the user accepts the terms of service

pwicherski:
I can confirm that on iOS we're sending FCM requests after accepting ToS 👍

advayDev1:
Ok so a test needs to be performed on this PR on both platforms that after you accept ToS and decline notifications that analytics still get sent.  Correct @patniemeyer ?

client/flutter/lib/pages/onboarding/onboarding_page.dart
```diff
@@ -47,14 +49,20 @@ class _OnboardingPageState extends State<OnboardingPage> {
...
+          LegalLandingPage(onNext: _onLegalDone),
           NotificationsPage(onNext: _toNextPage),
           LocationSharingPage(onNext: _onDone),
         ],
       ),
     );
   }
 
+  Future<void> _onLegalDone() async {
+    // Enable auto init so that analytics will work
```
advayDev1, line 61: Did we test without this?  Did analytics not get sent?

epicfaace:
> Ok so a test needs to be performed on this PR on both platforms that after you accept ToS and decline notifications that analytics still get sent. Correct @patniemeyer ?

Correct -- we only tested that the network calls were sent after accepting ToS, but did not actually test that analytics were sent. Will need to do that.

pwicherski:
@advayDev1 @patniemeyer I've tested it only on iOS, as in the description of the PR.

If we want to check those logs on Android device then we'll need to add HTTP logging to our app :)

We can use one of these solutions, but they all require changes in the code:
https://medium.com/flutter-community/three-easy-ways-to-debug-network-requests-in-flutter-53043e898929

patniemeyer:
@pwicherski It might be easiest to run it in the Android emulator and snoop the network on the host machine.  I am not currently in a position to test but will try tonight if I don't hear otherwise.

client/flutter/lib/pages/onboarding/onboarding_page.dart
epicfaace: Just realized that you were asking a separate question from what I commented below. No, we did not test without this -- I just assumed that FCM autoinit would be necessary for analytics to function.

This means that we need to test:
- with _onLegalDone -- make sure analytics are actually sent after accepting ToS
- without _onLegalDone -- see if analytics are still sent after accepting ToS

pwicherski:
@patniemeyer I've didn't managed to achieve this because currently, Flutter does not use the system-level proxy. I've tried to listen to the whole Qemu machine and requests from our app weren't there.

But if you could manage it when it would be awesome and I would love to know how :)

patniemeyer:
So I tried running the emulator with the `-tcpdump` option to capture all of its traffic, however I am not able to distinguish firebase calls from other random Android network calls from the OS since they all seem to use the same `1e100.net` domain name for lookup.   There is probably a more straightforward way to use the debugger to set a breakpoing on activity from the firebase code, but I am not familiar with Android Studio debugging.  I am inclined to simply accept Google's deferred init configuration as trustworthy, especially since it appears to work on iOS.  But if we really need to test this perhaps we can reach out to someone at Google for the quickest method?  @crazybob

epicfaace:
One thing we need to verify before merging this is that analytics still works (through Firebase DebugView) after going through the initial screen.

client/flutter/lib/pages/onboarding/onboarding_page.dart
advayDev1: Please retest this and get it submitted.

epicfaace:
@pwicherski tested this and noted that it crashes right after clicking the button to enable location sharing. Don't merge this for now

epicfaace:
> @pwicherski tested this and noted that it crashes right after clicking the button to enable location sharing. Don't merge this for now

never mind, this is no longer an issue — it worked after merging master into the PR branch

epicfaace:
We've verified that analytics does in fact work on this PR.

> This means that we need to test:
> 
> * with _onLegalDone -- make sure analytics are actually sent after accepting ToS
> * without _onLegalDone -- see if analytics are still sent after accepting ToS

@advayDev1 I tested, and whether or not we removed `_onLegalDone`, analytics are still properly sent. I think we should still keep it in, though, given that it is recommended in https://firebase.google.com/docs/cloud-messaging/android/client#prevent-auto-init.

PR #904, opened 2020-04-06
chore - update case stats

advayDev1:


server/appengine/src/main/java/who/WhoServiceImpl.java
```diff
@@ -25,17 +25,18 @@
...
     // TODO: Get from data source.
     return new GetCaseStatsResponse.Builder()
       .globalStats(new CaseStats.Builder()
-        .cases(719758L)
-        .deaths(33673L)
-        // 2020-03-31 10:00 CET
-        .lastUpdated(1585641600000L)
+        .cases(1136851L)
+        .deaths(62955L)
+        // 2020-05-04 18:00 CET
```
hspinks, line 38: nit: date format

```suggestion
        // 2020-04-05 18:00 CET
```

server/appengine/src/main/java/who/WhoServiceImpl.java
```diff
@@ -25,17 +25,18 @@
...
+        .lastUpdated(1588608000000L)
```
advayDev1, line 39:
```suggestion
        .lastUpdated(1586102400000L)
```

PR #921, opened 2020-04-07
Client-side caching of server content

patniemeyer:
This PR adds client side caching of the yaml files using the `flutter_cache_manager` package.  Cache expiration is controlled by the server cache control max-age response header (currently 10 minutes).  

I tested this by logging when the files were downloaded vs returned from the cache.  I also did a negative test by breaking the URL in order to see that the fallback behavior was still applied.

This PR adds two package: `https://pub.dev/packages/flutter_cache_manager` and  `https://pub.dev/packages/path_provider`.  The `path_provider` was already a (transitive) dependency of our project and is not new.

### DIFF
--- a/client/flutter/lib/api/content_loading.dart
+++ b/client/flutter/lib/api/content_loading.dart
@@ -4,7 +4,6 @@ import 'package:WHOFlutter/api/endpoints.dart';
 import 'package:WHOFlutter/components/dialogs.dart';
 import 'package:WHOFlutter/generated/l10n.dart';
 import 'package:flutter_cache_manager/flutter_cache_manager.dart';
-import 'package:http/http.dart' as http;
 import 'package:flutter/services.dart' show rootBundle;
 import 'package:flutter/material.dart';
 import 'package:path_provider/path_provider.dart';
@@ -89,7 +88,7 @@ class ContentLoading {
   /// Load a localized content bundle from the network, throwing an exception if not found.
   Future<ContentBundle> _loadFromNetwork(String name, String suffix) async {
     var url = '$baseContentURL/${_fileName(name, suffix)}';
-    var headers = {"Accept": "application/yaml"};
+    final headers = {"Accept": "application/yaml"};
     File file = await WhoCacheManager()
         .getSingleFile(url, headers: headers)
         .timeout(networkTimeout);
@@ -132,7 +131,7 @@ class WhoCacheManager extends BaseCacheManager {
   // expired file before returning, rather than favoring the cached version.
   @override
   Future<File> getSingleFile(String url, {Map<String, String> headers}) async {
-    var cacheFile = await getFileFromCache(url);
+    final cacheFile = await getFileFromCache(url);
     if (cacheFile != null) {
       if (cacheFile.validTill.isBefore(DateTime.now())) {
         // Wait for the refreshed file.
@@ -142,7 +141,7 @@ class WhoCacheManager extends BaseCacheManager {
     }
     try {
       print("Downloading file: $url");
-      var download = await webHelper.downloadFile(url, authHeaders: headers);
+      final download = await webHelper.downloadFile(url, authHeaders: headers);
       return download.file;
     } catch (e) {
       print("Error downloading file: $url, $e");
--- a/client/flutter/lib/api/content_loading.dart
+++ b/client/flutter/lib/api/content_loading.dart
@@ -1,11 +1,15 @@
+import 'dart:io';
 
 import 'package:WHOFlutter/api/endpoints.dart';
 import 'package:WHOFlutter/components/dialogs.dart';
 import 'package:WHOFlutter/generated/l10n.dart';
+import 'package:flutter_cache_manager/flutter_cache_manager.dart';
 import 'package:http/http.dart' as http;
 import 'package:flutter/services.dart' show rootBundle;
 import 'package:flutter/material.dart';
+import 'package:path_provider/path_provider.dart';
 import 'content_bundle.dart';
+import 'package:path/path.dart' as paths;
 
 class ContentLoading {
   static final networkLoadingEnabled = true;
@@ -85,12 +89,14 @@ class ContentLoading {
   /// Load a localized content bundle from the network, throwing an exception if not found.
   Future<ContentBundle> _loadFromNetwork(String name, String suffix) async {
     var url = '$baseContentURL/${_fileName(name, suffix)}';
-    var response = await http.get(url,
-        headers: {"Accept": "application/yaml"}).timeout(networkTimeout);
-    if (response.statusCode != 200) {
-      throw Exception("Error status code: ${response.statusCode}");
+    var headers = {"Accept": "application/yaml"};
+    File file = await WhoCacheManager()
+        .getSingleFile(url, headers: headers)
+        .timeout(networkTimeout);
+    if (file == null) {
+      throw Exception("File not retrieved from network or cache: $url");
     }
-    return ContentBundle.fromBytes(response.bodyBytes);
+    return ContentBundle.fromBytes(file.readAsBytesSync());
   }
 
   /// Load a localized content bundle from local assets, throwing an exception if not found.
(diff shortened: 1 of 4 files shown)


client/flutter/lib/api/content_loading.dart
```diff
@@ -85,12 +89,14 @@ class ContentLoading {
...
-      throw Exception("Error status code: ${response.statusCode}");
+    var headers = {"Accept": "application/yaml"};
+    File file = await WhoCacheManager()
+        .getSingleFile(url, headers: headers)
+        .timeout(networkTimeout);
+    if (file == null) {
+      throw Exception("File not retrieved from network or cache: $url");
     }
-    return ContentBundle.fromBytes(response.bodyBytes);
+    return ContentBundle.fromBytes(file.readAsBytesSync());
```
advayDev1, line 99: this sync call - does it hold up the ui thread ?

client/flutter/lib/api/content_loading.dart
```diff
@@ -105,3 +111,48 @@ class ContentLoading {
...
+
+  factory WhoCacheManager() {
+    if (_instance == null) {
+      _instance = WhoCacheManager._init();
+    }
+    return _instance;
+  }
+
+  // This changes the behavior of the base method to wait for an update of the
+  // expired file before returning, rather than favoring the cached version.
```
advayDev1, line 132: nice

client/flutter/lib/api/content_loading.dart
```diff
@@ -105,3 +111,48 @@ class ContentLoading {
...
+  @override
+  Future<File> getSingleFile(String url, {Map<String, String> headers}) async {
+    var cacheFile = await getFileFromCache(url);
```
advayDev1, line 135:
```suggestion
    final cacheFile = await getFileFromCache(url);
```

client/flutter/lib/api/content_loading.dart
```diff
@@ -105,3 +111,48 @@ class ContentLoading {
...
+    if (cacheFile != null) {
+      if (cacheFile.validTill.isBefore(DateTime.now())) {
+        // Wait for the refreshed file.
+        await webHelper.downloadFile(url, authHeaders: headers);
+      }
+      return cacheFile.file;
+    }
+    try {
+      print("Downloading file: $url");
+      var download = await webHelper.downloadFile(url, authHeaders: headers);
```
advayDev1, line 145:
```suggestion
      final download = await webHelper.downloadFile(url, authHeaders: headers);
```

client/flutter/lib/api/content_loading.dart
```diff
@@ -105,3 +111,48 @@ class ContentLoading {
...
+      return download.file;
+    } catch (e) {
+      print("Error downloading file: $url, $e");
+      return null;
+    }
+  }
+
+  // Store files in the temporary system path (default behavior)
+  Future<String> getFilePath() async {
+    var directory = await getTemporaryDirectory();
```
advayDev1, line 155: What are the security effects of doing this?

client/flutter/lib/api/content_loading.dart
```diff
@@ -85,12 +89,14 @@ class ContentLoading {
   /// Load a localized content bundle from the network, throwing an exception if not found.
   Future<ContentBundle> _loadFromNetwork(String name, String suffix) async {
     var url = '$baseContentURL/${_fileName(name, suffix)}';
-    var response = await http.get(url,
-        headers: {"Accept": "application/yaml"}).timeout(networkTimeout);
-    if (response.statusCode != 200) {
-      throw Exception("Error status code: ${response.statusCode}");
+    var headers = {"Accept": "application/yaml"};
```
advayDev1, line 92:
```suggestion
    final headers = {"Accept": "application/yaml"};
```

client/flutter/lib/api/content_loading.dart
```diff
@@ -105,3 +111,48 @@ class ContentLoading {
...
+
+  // This changes the behavior of the base method to wait for an update of the
+  // expired file before returning, rather than favoring the cached version.
+  @override
+  Future<File> getSingleFile(String url, {Map<String, String> headers}) async {
+    var cacheFile = await getFileFromCache(url);
+    if (cacheFile != null) {
+      if (cacheFile.validTill.isBefore(DateTime.now())) {
+        // Wait for the refreshed file.
+        await webHelper.downloadFile(url, authHeaders: headers);
```
advayDev1, line 139: I think we have a different requirement than normal HTTP caches.
If this call fails, we should still give the stale cached file.  A cached file is still better for the user than the baked in file, for WHO content files.

client/flutter/lib/api/content_loading.dart
```diff
@@ -1,11 +1,15 @@
+import 'dart:io';
 
 import 'package:WHOFlutter/api/endpoints.dart';
 import 'package:WHOFlutter/components/dialogs.dart';
 import 'package:WHOFlutter/generated/l10n.dart';
+import 'package:flutter_cache_manager/flutter_cache_manager.dart';
 import 'package:http/http.dart' as http;
```
advayDev1, line 7:
```suggestion
```

client/flutter/lib/api/content_loading.dart
patniemeyer: No, those are both called within async load methods (and they are also very fast anyway).

client/flutter/lib/api/content_loading.dart
patniemeyer: good point... fixing.

client/flutter/lib/api/content_loading.dart
patniemeyer: This returns a private, per-app temp storage area.  On iOS this uses NSCaches with the app's bundle id to get this location.  This package (path_provider) is maintained by the official flutter developer team.  This particular behavior is identical to the default caching package impl, I was simply required to repeat it here in the subclass.

client/flutter/lib/api/content_loading.dart
advayDev1: I don't think this is how async works.  Only if you use await do you allow the main event loop to keep running. 

Perhaps:
```suggestion
    return ContentBundle.fromBytes(await file.readAsBytes());
```

client/flutter/lib/api/content_loading.dart
epicfaace: getTemporaryDirectory [is analogous to](https://flutter.dev/docs/cookbook/persistence/reading-writing-files#1-find-the-correct-local-path):

> A temporary directory (cache) that the system can clear at any time. On iOS, this corresponds to the [NSCachesDirectory](https://developer.apple.com/documentation/foundation/nssearchpathdirectory/nscachesdirectory). On Android, this is the value that [getCacheDir()](https://developer.android.com/reference/android/content/Context#getCacheDir()) returns.

It appears that neither can be written to be other apps, although files in these directories can be deleted by the system / by the behavior of another app if there is a need to free up more storage space.

client/flutter/lib/api/content_loading.dart
patniemeyer: I've changed this.

client/flutter/lib/api/content_loading.dart
patniemeyer: Ah, sorry, I didn't even see what you highlighted (the sync method).  Yes, that is better... Although I don't think it would be a problem here.  Adding your suggestion!

advayDev1:
LG code-wise need to investigate all these deps being added.

patniemeyer:
@advayDev1 If the dependencies look too hair we could write our own caching very easily.  I was just trying to leverage what appeared to be a popular package.

advayDev1:
https://pub.dev/packages/synchronized - verified dev, 100/100
https://pub.dev/packages/flutter_cache_manager - verified dev, 99/100
sqflite - flutter favorite badge, verified dev
sqflite_common - unverified dev, but a dep of sqflite

the path providers are flutter official

Still looking into FMDB

PR #939, opened 2020-04-07
[feat] Store jurisdiction of client; closes #890

### LINKED ISSUE #890: Store political jurisdiction of client for location-specific notifications/content
We no longer use jittered GPS, only the city-center of the jurisdiction (down to locality, sub-administrative area, administrative area, and country).
Update the server accordingly.

advayDev1:


crazybob:
Should we use the Google Maps geocoder so we have consistency of results across platforms?

Should we store all of their fields (aside from the fine grained fields)? https://developers.google.com/maps/documentation/geocoding/intro#Types

crazybob:
Perhaps we could use the same structure as their address_components array. We would remove street_number and route.

advayDev1:
@crazybob PTAL.  Added country, adminArea1-5, and locality (city).  Think we should always store the short name/code and not tie our structure to GMaps' specific schema.

Not sure what we'd do with sub-city info and whether it remains non-identifiable at such detail.  From what I can tell sub-city info is not officially defined and can also be informal.

Also GMaps seems to have a 50QPS limit on geocoding; not sure if we can increase that.

crazybob:
> Also GMaps seems to have a 50QPS limit on geocoding; not sure if we can increase that.

I can ask for an increase.

crazybob:
We plan to perform geocoding in the client, right?

advayDev1:
> We plan to perform geocoding in the client, right?
@crazybob 

Yep, we'll move over to Google Maps in a coming PR on the client (we're using platform default right now).  I think you're def right about being consistent.

PR #949, opened 2020-04-08
fix: firebase app default does not exist warning handled

### LINKED ISSUE #948: Firebase implementation for iOS throws error: __FIRAPP_DEFAULT does not exist

When the application runs on the simulator, the following logs are displayed:

Configuring the default Firebase app...
6.21.0 - [Firebase/Core][I-COR000004] App with name __FIRAPP_DEFAULT does not exist.
Configured the default Firebase app __FIRAPP_DEFAULT.

stevan-milovanovic:
- [x] REQUIRED: Do you have an Issue **assigned to you** within the v1 milestone for this PR?  Put the Issue number here: #948

Configure a FirebaseApp shared instance

Added line uses Firebase library to configure APIs

Additionally, unnecessary cast from the line 26 was removed.

closes #948

No

After the changes, error mentioned in the issue #948 could not be reproduced.

### DIFF
--- a/client/flutter/ios/Runner/AppDelegate.swift
+++ b/client/flutter/ios/Runner/AppDelegate.swift
@@ -23,7 +23,7 @@ import Firebase
     }
 
     if #available(iOS 10.0, *) {
-      UNUserNotificationCenter.current().delegate = self
+      UNUserNotificationCenter.current().delegate = self as? UNUserNotificationCenterDelegate
     }
 
     return super.application(application, didFinishLaunchingWithOptions: launchOptions)
--- a/client/flutter/ios/Runner/AppDelegate.swift
+++ b/client/flutter/ios/Runner/AppDelegate.swift
@@ -1,5 +1,6 @@
 import UIKit
 import Flutter
+import Firebase
 
 @UIApplicationMain
 @objc class AppDelegate: FlutterAppDelegate {
@@ -7,6 +8,7 @@ import Flutter
     _ application: UIApplication,
     didFinishLaunchingWithOptions launchOptions: [UIApplication.LaunchOptionsKey: Any]?
   ) -> Bool {
+    FirebaseApp.configure()
     GeneratedPluginRegistrant.register(with: self)
     // Work-around for https://github.com/flutter/flutter/issues/39032
     // This is safe to remove once the fix for that is available in the SDK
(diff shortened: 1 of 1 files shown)


advayDev1:
I'd like to understand more before this gets merged:

What is the impact of the error?  Why does this matter?

What is the downside (if any) of configuring the FirebaseApp here?

Will the PR cause any outgoing HTTP requests to get sent on app start, possibly before users have been onboarded?

Does the Android code need a similar fix?

client/flutter/ios/Runner/AppDelegate.swift
```diff
@@ -21,7 +23,7 @@ import Flutter
     }
 
     if #available(iOS 10.0, *) {
-      UNUserNotificationCenter.current().delegate = self as? UNUserNotificationCenterDelegate
+      UNUserNotificationCenter.current().delegate = self
```
epicfaace, line 26: Can we change this back to the default? I know it's unnecessary but it's good practice to keep only the changes that fix this particular issue to this PR and do other refactoring in subsequent PRs.
```suggestion
      UNUserNotificationCenter.current().delegate = self as? UNUserNotificationCenterDelegate
```

epicfaace:
See https://github.com/flutter/flutter/issues/16871 -- it appears this issue can cause the app to crash on iOS. Though the solution of using `FirebaseApp.configure()` is detailed in the aforementioned issue, it's not mentioned in the [official Firebase + Flutter setup guide](https://firebase.google.com/docs/flutter/setup?platform=ios) (though it is mentioned in the [official Firebase setup guide for iOS](https://firebase.google.com/docs/ios/setup)).

@stevan-milovanovic -- to confirm, when you ran the app, did it crash, or did it just show this error in the console?

stevan-milovanovic:
@advayDev1, as @epicfaace already mentioned, this could potentially cause the app to crash on iOS. 
I do not see downsides since this is from the official guide for integrating Firebase for iOS projects.
Unfortunately, I could not test this point (outgoing HTTP requests) because Firebase could not be used on the simulator and I am currently without a real test device.
I will have to check for the Android.

@epicfaace the error is only displayed in the console.

epicfaace:
@stevan-milovanovic right, but did you actually see the app crash when you tried it? It would be good to know what exactly / when it will crash before merging it.

The main downside I see is that this is not from the official guide for Firebase integration with Flutter -- it's only for the Firebase integration with iOS (which isn't designed for Flutter apps).

If we can reproduce the crash at least once, then I think it would be good to merge this, though.

stevan-milovanovic:
@epicfaace, I haven't seen the crash of the app, just the warning in the console. But since I found this issue on the official flutter repository (https://github.com/flutter/flutter/issues/30765), I supposed that it would be helpful to exclude the possibility of crashing.

PR #962, opened 2020-04-08
Latest Numbers timeseries data struct

advayDev1:
#961

### DIFF
--- a/api/who/who.proto
+++ b/api/who/who.proto
@@ -43,6 +43,18 @@ message GetCaseStatsResponse {
   uint64 ttl = 2;
 }
 
+message StatSnapshot {
+  uint64 epochMsec = 1;
+
+  int64 dailyCases = 2;
+  int64 dailyDeaths = 3;
+  int64 dailyRecoveries = 4;
+
+  int64 totalCases = 5;
+  int64 totalDeaths = 6;
+  int64 totalRecoveries = 7;
+}
+
 // Statistics in a single or aggregate jurisdiction.
 message CaseStats {
   JurisdictionType jurisdictionType = 7;
@@ -57,6 +69,8 @@ message CaseStats {
   int64 recoveries = 5;
 
   string attribution = 6;
+
+  StatSnapshot[] timeseries = 8;
 }
 
 enum JurisdictionType {


api/who/who.proto
```diff
@@ -43,6 +43,18 @@ message GetCaseStatsResponse {
   uint64 ttl = 2;
 }
 
+message StatSnapshot {
+  uint64 epochMsec = 1;
```
hspinks, line 47: in the array of StatSnapshots, can I assume that there will be 1 StatSnapshot per 24 hour period (e.g. each epoch is exactly 24 hours apart), or will the client have to aggregate?

advayDev1: I think the client should not make that assumption.  The client can control its histogram buckets, so it should do that at any granularity.

hspinks: will that yield accurate counts though? e.g. how does the underlying data source count its daily cases?

hspinks: The concern is: what if the daily counts are recorded at inconsistent times during the day, but intended to record that day's "official" count. Let's say that one day, the daily count is recorded at 10pm and the next day, the daily count is recorded at 9pm. If someone accesses the app in a timezone 3 hours ahead of where the daily counts are recorded, then those 2 StatSnapshots may be aggregated into 1 day, and the previous day would not have any StatSnapshots

advayDev1: There's no documentation ofc ;). But it appears the underlying data has exactly 0 or 1 entry for each 24hr period *ending* at midnight GMT.

For example US at 1586304000000 (April 8, 2020 0000hr GMT) has 29,510 deaths which is shown as "April 8, 2020" even though April 8th hasn't ended yet here.

hspinks: Ok, so the client should probably align the days with GMT rather than local time zone?

api/who/who.proto
```diff
@@ -57,6 +69,8 @@ message CaseStats {
   int64 recoveries = 5;
 
   string attribution = 6;
+
+  StatSnapshot[] timeseries = 8;
```
hspinks, line 73: confirming that this will respond with every single snapshot associated with the requested jurisdiction? no limit on the number of snapshots / no paging mechanism / no way to request specific days, correct?

advayDev1: Correct.  Don't think we should make it complicated right now.

api/who/who.proto
advayDev1: Given we have no API docs, I'd say yes

PR #963, opened 2020-04-08
Run all tests when apis change

advayDev1:


.github/workflows/flutter-build.yml
```diff
@@ -3,11 +3,13 @@ on:
   push:
     paths:
       - '.github/workflows/**'
+      - 'api/**'
```
hspinks, line 6: will the Flutter build depend on API changes?

.github/workflows/flutter-test.yml
```diff
@@ -3,11 +3,13 @@ on:
   push:
     paths:
       - '.github/workflows/**'
+      - 'api/**'
```
hspinks, line 6: Same question... if our Flutter tests are unit tests and not integration tests, they wouldn't depend on API changes. I think there's a chance of false positive tests that give us a false sense of security (e.g. the API could change, we don't update our Flutter tests to the new API spec, Flutter unit tests still pass, but app runs fail).

Unless we think that we'll be writing Flutter tests that actually hit the APIs, but that would be a different conversation...

.github/workflows/flutter-build.yml
advayDev1: The API directory defines both the service and also the data structures.  The Flutter client hand-rolls the JSON but it should use for example protoc-generated dart classes: https://developers.google.com/protocol-buffers/docs/darttutorial

.github/workflows/flutter-test.yml
advayDev1: Its not about hitting the APIs.  Its that the client should compile based on protoc-generated dart classes from the protobufs in api/

PR #977, opened 2020-04-09
Add next fact button to carousel view

theswerd:

Adds **Next fact** button to the carousel view

- Refactors carousel slide into new file
- Adds all carousel components to carousel folder

No

iOS simulator

Simulator Screen Shot - iPhone 11 Pro Max - 2020-04-09 at 12 09 55


patniemeyer:
A few things: 0) Thanks for reorganizing the files - Nice.  1) When you reach the last item the button still tries to push you forward.  2) The next fact button FlatButton should probably be wrapped in SafeArea.  Right now it sits very close to the page indicator on iOS.

image

PR #982, opened 2020-04-10
[feat] Add user agent and accept gzip encoding, closes #976

### LINKED ISSUE #976: Include "gzip" and app version in User-Agent for all HTTP requests from client
The "User-Agent" header is currently set to "Dart/2.7 (dart:io)".

Set it on all HTTP requests to: `User-Agent: WHO-App/[VERSION] (gzip)`

Including "gzip" in the User-Agent tells Google's frontend that it can gzip responses.

Examples:

- RPCs
- YAML content updates
- Others?

advayDev1:
fixes #976

### DIFF
--- a/client/flutter/lib/api/who_service.dart
+++ b/client/flutter/lib/api/who_service.dart
@@ -1,4 +1,5 @@
 import 'dart:convert';
+import 'package:WHOFlutter/main.dart';
 import 'package:WHOFlutter/api/endpoints.dart';
 import 'package:WHOFlutter/api/user_preferences.dart';
 import 'package:http/http.dart' as http;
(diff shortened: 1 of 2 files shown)


client/flutter/lib/api/who_service.dart
```diff
@@ -57,11 +58,16 @@ class WhoService {
...
       'Who-Client-ID': clientId,
-      'Who-Platform': _platform
+      'Who-Platform': _platform,
+      'User-Agent': userAgent,
     };
     return headers;
   }
 
+  static String get userAgent {
+    return 'WHO-App/${_platform}/${packageInfo != null ? packageInfo.version : ''}/${packageInfo != null ? packageInfo.buildNumber : ''} (gzip)';
```
epicfaace, line 68: When would `packageInfo` be equal to `null`?

advayDev1: On web https://github.com/WorldHealthOrganization/app/blob/master/client/flutter/lib/main.dart#L27

PR #1022, opened 2020-04-13
Dynamic "Protect Yourself" and "Travel Advice" pages plus refactoring of content bundle mechanism...

patniemeyer:
This PR changes the "Protect Yourself" and "Travel Advice" pages to dynamic content along with significant refactoring of the content bundle API. 

1) "Protect Yourself" is now driven by a "fact" schema, which has been extended to support an `animation_name` in addition to `image_name`.
2) "Travel Advice" is now driven by a new "advice" schema.  This schema includes additional, fixed values (e.g. page banner text) that are managed outside of the items list and can be extended. 
3) `ContentBundle` now supports an `unsupportedSchemaVersionAvailable` flag that can be handled by the calling page rather than popping the dialog within the content loading mechanism.  The dialog has moved to a helper method in the `Dialogs` class.    
4) The content loading API no longer requires `BuildContext` and accepts a `Locale` instead.
5) Content is interpreted by individual classes for each schema, see `api/content/schema/*` and these are used by the calling classes directly.

I will continue testing this but wanted to get a review started.


client/flutter/lib/pages/travel_advice.dart
```diff
@@ -1,85 +1,130 @@
...
+          description: _adviceContent?.recommendations ?? "",
+          verticalPadding: 16,
+          titleStyle: TextStyle(fontSize: 16, fontWeight: FontWeight.bold),
+          descriptionColor: Colors.white,
+        ),
+      )
+    ]));
+  }
+
+  SliverToBoxAdapter _buidLoading() {
```
advayDev1, line 107:
```suggestion
  SliverToBoxAdapter _buildLoading() {
```

client/flutter/lib/pages/travel_advice.dart
```diff
@@ -1,85 +1,130 @@
...
+      ..._getItems(context),
+      SizedBox(
+        height: 30,
+      ),
+      Padding(
+        padding: const EdgeInsets.symmetric(horizontal: 8.0, vertical: 10.0),
+        child: PageButton(
+          Color(0xff008DC9),
+          S.of(context).travelAdvicePageButtonGeneralRecommendations,
+          () => launch(
```
advayDev1, line 96: If there's no link the button should not exist IMO.

client/flutter/lib/pages/travel_advice.dart
```diff
@@ -1,85 +1,130 @@
...
   }
+
+  SliverList _buildBody(BuildContext context) {
+    return SliverList(
+        delegate: SliverChildListDelegate([
+      Container(
+        color: Color(0xffD82037),
+        padding: EdgeInsets.symmetric(horizontal: 25, vertical: 10),
+        child: Html(
+          //data: S.of(context).travelAdviceContainerText,
```
advayDev1, line 63: Delet

client/flutter/lib/pages/travel_advice.dart
patniemeyer: Fixed

client/flutter/lib/pages/travel_advice.dart
patniemeyer: Fixed

client/flutter/lib/pages/travel_advice.dart
patniemeyer: Fixed

client/flutter/assets/content_bundles/protect_yourself.en.yaml
```diff
@@ -0,0 +1,26 @@
...
+    -
+      image_name: elbow
+      body_html: |
+        *Cover your mouth and nose* with your bent elbow or tissue when you cough or sneeze.
+    -
+      image_name: social_distance
+      body_html: |
+        *Stay more than* 1 meter (>3 feet) away from a person who is sick.
+    -
+      image_name: mask
```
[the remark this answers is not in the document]
patniemeyer: @hspinks Worth noting here also that the `protect_yourself` yaml contains plain text with `*` asterisks for bold rather than our regular HTML content.  This was to maintain consistency with the current page implementation but can be changed if desired.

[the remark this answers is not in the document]
advayDev1: Think we should use just restricted html going forward and not roll our own less-than-markdown.  BUt its for a future PR :D

client/flutter/lib/api/content/content_bundle.dart
```diff
@@ -43,17 +53,46 @@ class ContentBundle {
...
     } catch (err) {
       return YamlList();
     }
   }
 
-  int _getInt(String key, {int orDefault = -1}) {
-    return yaml[key];
+  String getString(String key) {
+    try {
+      return (yaml['contents'][key]).trim();
```
advayDev1, line 66: Why would we trim a string?

client/flutter/lib/api/content/schema/advice_content.dart
```diff
@@ -0,0 +1,51 @@
...
+    var bundle = await ContentLoading().load(locale, "travel_advice");
+    return AdviceContent(bundle);
+  }
+
+  AdviceContent(ContentBundle bundle) : super(bundle, schemaName: "advice") {
+    try {
+      this.banner = bundle.getString('banner_html') ?? "";
+      this.recommendations = bundle.getString('recommendations_html') ?? "";
+      this.recommendationsLink =
+          bundle.getString('recommendations_link') ?? "https://www.who.int";
```
advayDev1, line 26: if no link, should not show link here either

client/flutter/lib/pages/travel_advice.dart
```diff
@@ -1,85 +1,132 @@
...
+        child: Padding(
+      padding: const EdgeInsets.all(48.0),
+      child: CupertinoActivityIndicator(),
+    ));
+  }
+
+  List<Widget> _getItems(BuildContext context) {
+    return (_adviceContent?.items ?? []).map((item) {
+      return TravelAdviceListItem(
+          imageSrc: 'assets/travel_advice/${item.imageName}.png',
```
advayDev1, line 120: File an issue to scrub all variables that go into path components.

They should have solely [a-zA-Z0-9/].

hspinks: #1028

client/flutter/assets/content_bundles/protect_yourself.en.yaml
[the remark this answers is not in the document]
hspinks: #1029 #1030

client/flutter/lib/api/content/schema/advice_content.dart
hspinks: #1031

client/flutter/lib/api/content/content_bundle.dart
advayDev1: Dont think this was resolved pre-merge @hspinks @patniemeyer .  Would still like to understand why we don’t leave it untrimmed.

patniemeyer: This is related to the fact that the Protect Yourself page is currently using plain text rather than HTML - The string from the YAML includes a trailing newline and that was breaking the layout in the current page implementation.  I decided that any YAML leading and trailing whitespace was just an artifact of the file format and so trimmed it.   If this is better addressed in the YAML somehow or if we treat everything as HTML (or trim on a case by case basis in the UI layer) then this can be removed.

PR #1023, opened 2020-04-13
[feat] DRAFT - Subscribe clients to topics based on locations

### LINKED ISSUE #1000: Allow Efficient targeting of notifications to whole countries
Current notification system does not efficiently allow us to target large regions for notifications.  We would have to run a query against Clients' location fields and then send batch notifications programmatically (which can only be sent approximately 1000 at a time).

Instead, we should use the feature built for such purpose: FCM topics.  On both `/putDeviceToken` AND `/putLocation`, we should subscribe the given FCM token for topics if both a token and location are at that time available.  The suggested topics to register for are:
```
LOCATION/$country
LOCATION/$country/$adminArea1 if adminArea1 not empty
LOCATION/$country/$adminArea1/$adminArea2 if adminArea2 not empty
LOCATION/$country/$adminArea1/$adminArea2/$adminArea3 if adminArea3 not empty
LOCATION/$country/$adminArea1/$adminArea2/$adminArea3/$adminArea4 if adminArea4 not empty
LOCATION/$country/$adminArea1/$adminArea2/$adminArea3/$adminArea4/$adminArea5 if adminArea5 not empty
LOCATION/$country/$adminArea1/$adminArea2/$adminArea3/$adminArea4/$adminArea5/$locality if locality not empty
```

* Note we must scrub these location fields as they are controlled by the user.  Any forward slashes in the names should be encoded as something else.
* The `LOCATION/` prefix gives us some namespaces for future non-location based topics.
* When `/putLocation` is called when a prior location exists in datastore, the client must be unregistered from its old location-based topics.
* App version and platform-based targeting are built-in to FCM.

cc @cbgrey @crazybob

advayDev1:
Closes #1000

### DIFF
--- a/server/appengine/src/main/java/who/FirebaseModule.java
+++ b/server/appengine/src/main/java/who/FirebaseModule.java
@@ -1,0 +1,34 @@
+package who;
+
+import com.google.auth.oauth2.GoogleCredentials;
+import com.google.firebase.FirebaseApp;
+import com.google.firebase.FirebaseOptions;
+import com.google.firebase.messaging.FirebaseMessaging;
+import com.google.inject.AbstractModule;
+import com.google.inject.Provides;
+
+import java.io.IOException;
+
+import javax.inject.Singleton;
+
+public class FirebaseModule extends AbstractModule {
+
+  @Provides @Singleton
+  FirebaseApp provideFirebaseApp() throws IOException {
+
+    FirebaseOptions options = new FirebaseOptions.Builder()
+      .setCredentials(GoogleCredentials.getApplicationDefault())
+      // FIXME
+      .setDatabaseUrl("https://who-myhealth-staging.firebaseio.com")
+      .build();
+
+    FirebaseApp.initializeApp(options);
+
+    return FirebaseApp.getInstance();
+  }
+
+  @Provides @Singleton
+  FirebaseMessaging provideFirebaseMessaging(FirebaseApp app) {
+    return FirebaseMessaging.getInstance(app);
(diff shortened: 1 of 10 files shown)


server/appengine/src/test/java/who/TopicsTest.java
```diff
@@ -0,0 +1,28 @@
...
+  }
+
+  @Test
+  public void testTopicNames() throws Exception {
+    Client c = new Client();
+    c.countryCode = "US";
+    c.adminArea1 = "CA";
+    c.adminArea2 = "San Francisco County";
+    c.locality = "San Francisco";
+    System.err.println(Topics.getTopicNames(c));
```
advayDev1, line 26: Put an actual test

server/appengine/src/main/java/who/FirebaseModule.java
```diff
@@ -0,0 +1,34 @@
...
+
+public class FirebaseModule extends AbstractModule {
+
+  @Provides @Singleton
+  FirebaseApp provideFirebaseApp() throws IOException {
+
+    FirebaseOptions options = new FirebaseOptions.Builder()
+      .setCredentials(GoogleCredentials.getApplicationDefault())
+      // FIXME
+      .setDatabaseUrl("https://who-myhealth-staging.firebaseio.com")
```
advayDev1, line 22: Get from environment

advayDev1:
No longer a v1 feature

advayDev1:
Submitted as a different PR

PR #1027, opened 2020-04-14
generate l10n code via CI / cmd

kassim:
- [x] REQUIRED: Do you have an Issue **assigned to you** within the v1 milestone for this PR?  Put the Issue number here: https://github.com/WorldHealthOrganization/app/issues/1019
- [ ] Included any relevant screenshots of UI updates.

Adds the `intl_utils` plugin to the pubspec file and add the generate command to the CI build, and documentation to inform other devs about the plugin. Also removes generated code from the build.

Closes https://github.com/WorldHealthOrganization/app/issues/1019

intl_utils: ^1.2.2

run `flutter pub run intl_utils:generate` and the same code is generated as was previously by the IDE plugins

## Discussion

- Perhaps a better place for the documentation is in the client/flutter directory?


.gitignore
```diff
@@ -96,3 +96,5 @@ buck-out/
 
 # Bundle artifact
 *.jsbundle
+
+client/flutter/lib/generated
```
hspinks, line 100: I think that in general, it makes sense to ignore the generated files since they can easily be regenerated. But, this will add an additional step to the build process when building for release. Eventually, build-for-release should be automated with fastlane, but until then, I'm wary of changing the process since we've established it well with WHO so far.

As an interim step, what would we think about keeping the generated files in the repo, but still accepting this change to add the l10n generation in our CI? That way, the CI will still fail on this line if a mistake is made:
 https://github.com/WorldHealthOrganization/app/blob/c44b7de0cbb511097aedde398c44d02f9eb0a2d6/.github/workflows/flutter-build.yml#L34

cc @advayDev1

client/flutter/pubspec.yaml
```diff
@@ -37,6 +37,7 @@ dependencies:
   fl_chart: ^0.8.7
 
   intl: ^0.16.0
+  intl_utils: ^1.2.2
```
patniemeyer, line 40: Since this is only used by the build process I think this dependency can be moved under `dev_dependencies` in the `pubspec.yaml`.

.gitignore
advayDev1: I agree with you Hunter on both points.

kassim:
so there's a lovely issue in that there is different ordering to the locales depending on which OS you generate the code with, meaning the git diff test will never pass for both client builds 😅 

a fix doesn't seem clear for me from our side, but I've filed an issue to Localizely https://github.com/localizely/flutter-intl-plugin-sample-app/issues/4, so maybe we keep this on hold for the minute

kassim:
@hspinks @advayDev1 this is building correctly now

i created a script to run the code generation + the format command so that the new tests would pass - there doesn't seem to be a simple way to ignore the generated code when running 'flutter format', ideally generated code wouldn't be in the source code directory but that can be addressed later (if possible at all)

.github/workflows/client-build.yaml
```diff
@@ -36,6 +38,8 @@ jobs:
       - run: echo "::add-path::$GITHUB_WORKSPACE/_flutter/bin"
       - run: flutter pub get
         working-directory: ./client/flutter
+      - run: sh ./helper-scripts/generate-l10n-code.sh
```
advayDev1, line 41: no need to put this in build when it's already in test

.github/workflows/client-build.yaml
```diff
@@ -21,6 +21,8 @@ jobs:
...
```
advayDev1, line 24: no need to put this in build when it's already in test

kassim:
cleaned up and merged in master

PR #1074, opened 2020-04-19
Update issue templates

advayDev1:


hspinks:
@advayDev1 why did you close this? Multiple templates seems like a good idea to me

advayDev1:
@hspinks I agree but I opened it prematurely.  I'll do it right and send a new one :D

PR #1085, opened 2020-04-20
Add britannio to flutter committers.

advayDev1:


advayDev1:
Thanks @britannio for taking on this additional responsibility of protecting the codebase and our users.

PR #1088, opened 2020-04-21
Change "Protect Yourself" page to use HTML for dynamic content.

### LINKED ISSUE #1030: Use HTML instead of markdown for Protect Yourself yaml
See https://github.com/WorldHealthOrganization/app/pull/1022#discussion_r407800233

---

patniemeyer:
This PR changes the "Protect Yourself" page to use HTML for the dynamic content instead of the ad-hoc markdown syntax.  

I believe it looks identical with the exception that one line is breaking slightly differently.

image

Closes #1030

### DIFF
--- a/client/flutter/lib/pages/protect_yourself.dart
+++ b/client/flutter/lib/pages/protect_yourself.dart
@@ -1,3 +1,4 @@
+import 'package:flutter_html/flutter_html.dart';
 import 'package:who_app/api/content/schema/fact_content.dart';
 import 'package:who_app/components/dialogs.dart';
 import 'package:who_app/components/page_scaffold/page_scaffold.dart';
@@ -6,6 +7,7 @@ import 'package:who_app/constants.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter_svg/flutter_svg.dart';
 import 'package:who_app/generated/l10n.dart';
+import 'package:html/dom.dart' as dom;
 
 class ProtectYourself extends StatefulWidget {
   final FactsDataSource dataSource;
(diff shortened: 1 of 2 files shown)


client/flutter/lib/pages/protect_yourself.dart
```diff
@@ -55,10 +57,26 @@ class _ProtectYourselfState extends State<ProtectYourself> {
...
+      color: Constants.textColor,
+      fontSize: 16 * MediaQuery.textScaleFactorOf(context),
+      height: 1.4,
+    );
+    final TextStyle boldText = normalText.copyWith(fontWeight: FontWeight.w700);
     return (_factContent?.items ?? []).map((fact) {
-      print("fact body = ${fact.body}, image = ${fact.imageName}");
       return _ProtectYourselfCard(
-        message: _message(fact.body),
+        message: Html(
```
epicfaace, line 68: How will this handle non-`<b>` tags? (I think the ideal is that non-`<b>` tags are not rendered)

patniemeyer: The Html widget has defaults for many of the standard tags.  We can customize their appearance in the same way that we do for the `<b>` tag in the `customTextStyle`.  I guess if we had a need to "turn off" tags we could also do it there.

epicfaace: I think it would be best to "turn off" all other tags if possible (other than a predefined set of tags) so we don't get any other unexpected behavior?

advayDev1: FYI to myself: Looks like a UI concern rather than a security one given `a, abbr, acronym, address, article, aside, b, bdi, bdo, big, blockquote, body, br, caption, cite, code, data, dd, del, dfn, div, dl, dt, em, figcaption, figure, footer, h1, h2, h3, h4, h5, h6, header, hr, i, img, ins, kbd, li, main, mark, nav, noscript, ol, p, pre, q, rp, rt, ruby, s, samp, section, small, span, strike, strong, sub, sup, table, tbody, td, template, tfoot, th, thead, time, tr, tt, u, ul, var` therefore, I don't think we should put in that effort.  

The risk here is a ugly UI from content we approved to go out.

client/flutter/lib/pages/protect_yourself.dart
```diff
@@ -55,10 +57,26 @@ class _ProtectYourselfState extends State<ProtectYourself> {
...
+            data: fact.body ?? "",
```
advayDev1, line 69: since this variable is used as html, rename to fact.bodyHtml

patniemeyer: We have 7 or 8 other fields in those `Item` classes that all contain HTML.  Do you want to  adopt that naming convention for them all now? (I'm not sure that we are consistent with this convention throughout the rest of the code).

purpledrosophila:
Hi, just checking in, is it safe to add my protect_yourself.en.yaml html from the previous issue to my pull request for the other YAML in the folder now? I was just going to create the pull request, noticed this issue was up.

PR #1089, opened 2020-04-21
Add loading indicator to "Protect Yourself" page and refactor widget.

patniemeyer:
This PR adds a loading indicator to the protect yourself pages and refactors the widget so that it is used in the same way across pages.

image

### DIFF
--- a/client/flutter/lib/components/loading_indicator.dart
+++ b/client/flutter/lib/components/loading_indicator.dart
@@ -10,9 +10,8 @@ class LoadingIndicator extends StatelessWidget {
   Widget build(BuildContext context) {
     return SliverToBoxAdapter(
         child: Padding(
-          padding: const EdgeInsets.only(top: 48.0),
-          child: CupertinoActivityIndicator(),
-        ));
+      padding: const EdgeInsets.only(top: 48.0),
+      child: CupertinoActivityIndicator(),
+    ));
   }
 }
-
--- a/client/flutter/lib/components/loading_indicator.dart
+++ b/client/flutter/lib/components/loading_indicator.dart
@@ -1,0 +1,18 @@
+import 'package:flutter/cupertino.dart';
+import 'package:flutter/material.dart';
+
+class LoadingIndicator extends StatelessWidget {
+  const LoadingIndicator({
+    Key key,
+  }) : super(key: key);
+
+  @override
+  Widget build(BuildContext context) {
+    return SliverToBoxAdapter(
+        child: Padding(
+          padding: const EdgeInsets.all(48.0),
+          child: CupertinoActivityIndicator(),
+        ));
+  }
+}
+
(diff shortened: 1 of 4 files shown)


client/flutter/lib/components/loading_indicator.dart
```diff
@@ -0,0 +1,18 @@
...
+class LoadingIndicator extends StatelessWidget {
+  const LoadingIndicator({
+    Key key,
+  }) : super(key: key);
+
+  @override
+  Widget build(BuildContext context) {
+    return SliverToBoxAdapter(
+        child: Padding(
+          padding: const EdgeInsets.all(48.0),
```
theswerd, line 13: This should only be top

client/flutter/lib/pages/protect_yourself.dart
```diff
@@ -48,12 +49,16 @@ class _ProtectYourselfState extends State<ProtectYourself> {
...
-        body: [SliverList(delegate: SliverChildListDelegate(_buildCards()))]);
+      title: S.of(context).protectYourselfTitle,
+      showShareBottomBar: false,
+      announceRouteManually: true,
+      body: [_factContent != null ? _buildBody() : LoadingIndicator()],
+    );
   }
 
+  SliverList _buildBody() =>
+      SliverList(delegate: SliverChildListDelegate(_buildCards()));
```
theswerd, line 60: Please add trailing commas

client/flutter/lib/components/loading_indicator.dart
```diff
@@ -0,0 +1,18 @@
+import 'package:flutter/cupertino.dart';
+import 'package:flutter/material.dart';
+
+class LoadingIndicator extends StatelessWidget {
```
britannio, line 4: Since this produces a Sliver,  `SliverLoadingIndicator` would be more fitting.

client/flutter/lib/pages/protect_yourself.dart
patniemeyer: What do you mean exactly?

client/flutter/lib/components/loading_indicator.dart
patniemeyer: Fixed.

client/flutter/lib/components/loading_indicator.dart
patniemeyer: In this context I think of it as *our* loading indicator for the app, not a particular style of loading indicator.  We may change its impl or styling from here...

client/flutter/lib/pages/protect_yourself.dart
theswerd:
```suggestion
      SliverList(
          delegate: SliverChildListDelegate(
               _buildCards(),
           ),
      );
```

patniemeyer: Maybe mean for the array in page scaffold that uses this...  Yah.  Wondering if that should be a 'children' and a 'body' alternative that takes a single child.

patniemeyer: Fixed (the commas)

PR #1124, opened 2020-04-25
Pin notification_permissions to v0.4.4

advayDev1:
Have looked at https://storage.googleapis.com/pub-packages/packages/notification_permissions-0.4.4.tar.gz


hspinks:
@advayDev1 this is fine with me, but what's the rationale?

advayDev1:
@hspinks I didn't trust this dev. But that's not objective.  Let's think through a pinning policy. Closing for now.

PR #1127, opened 2020-04-25
feat: Update settings page with new menu; closes #1115

### LINKED ISSUE #1115: Add bottom menu to Settings page

<details>
<summary>Rough mock (subject to change)</summary>
Screen Shot 2020-04-22 at 6 11 00 PM
</details>

advayDev1:
closes #1115

Screen Shot 2020-04-25 at 5 02 39 PM


client/flutter/lib/pages/settings_page.dart
```diff
@@ -127,6 +133,45 @@ class _SettingsPageState extends State<SettingsPage>
                         .homePagePageSliverListSettingsNotificationsInfo,
                     isToggled: _notificationsEnabled ?? false,
                     onToggle: _toggleNotifications),
+                SizedBox(
+                  height: 20,
+                ),
+                divider,
+                menuItem(
```
hspinks, line 140: Not a blocker for this PR, but in the model of the UI of the page, it feels a little weird to have individual menu items as children of the main column of the page as opposed to a menu container. It's not worth making this a fully separate widget file, but it might be worth nesting those items. But, it still does work the way it is, so this is not necessary to merge this.

client/flutter/lib/pages/settings_page.dart
```diff
@@ -127,6 +133,45 @@ class _SettingsPageState extends State<SettingsPage>
...
+                      FirebaseAnalytics().logEvent(name: 'Feedback');
+                      // TODO: Implement feedback #989 #1015
+                    }),
+                divider,
+                menuItem(
+                  title: S.of(context).homePagePageSliverListAboutTheApp,
+                  onTap: () {
+                    FirebaseAnalytics().logEvent(name: 'About');
+                    return Navigator.of(context)
+                        .push(MaterialPageRoute(builder: (c) => AboutPage()));
```
hspinks, line 168: Are we still using `MaterialPageRoute` in the Cupertino app? Do we want to consider switching to whatever new routing / linking scheme we end up using for the app here, too?

client/flutter/lib/pages/settings_page.dart
```diff
@@ -127,6 +133,45 @@ class _SettingsPageState extends State<SettingsPage>
...
+                  },
+                ),
+                divider,
+                Container(
+                  height: 25,
```
hspinks, line 173: The nitpickiest nit that ever did pick: David requested all gutters be sizes in multiples of 4 until we have a more coherent way of handling padding. Not necessary if you don't want to, but:

```suggestion
                  height: 24,
```

advayDev1: Done.  However David's paddings are not currently multiples of 4 in his figmas, so won't touch the others.

client/flutter/lib/pages/settings_page.dart
advayDev1: Eventually yes, but I think for now this gives us a platform-comfortable feel.  We can change all of them once David makes a decision.

client/flutter/lib/pages/settings_page.dart
advayDev1: Good point and fixed

PR #1129, opened 2020-04-25
Improve travel advice page; closes #1128

### LINKED ISSUE #1128: Refresh Travel Advice page

advayDev1:
closes #1128

Will need pixel perfect eventually, 

font awesome dependency, via flutter community official acct

Screenshot:
Screen Shot 2020-04-25 at 1 44 06 AM

Mock:
Screen Shot 2020-04-24 at 11 45 56 PM

### DIFF
--- a/client/flutter/lib/pages/travel_advice.dart
+++ b/client/flutter/lib/pages/travel_advice.dart
@@ -1,12 +1,11 @@
+import 'package:font_awesome_flutter/font_awesome_flutter.dart';
 import 'package:who_app/api/content/schema/advice_content.dart';
 import 'package:who_app/components/dialogs.dart';
-import 'package:who_app/components/page_button.dart';
 import 'package:who_app/components/page_scaffold/page_scaffold.dart';
+import 'package:who_app/constants.dart';
 import 'package:who_app/generated/l10n.dart';
 import 'package:flutter/cupertino.dart';
 import 'package:flutter/material.dart';
-import 'package:url_launcher/url_launcher.dart';
-import 'package:flutter_html/flutter_html.dart';
 
 class TravelAdvice extends StatefulWidget {
   final AdviceDataSource dataSource;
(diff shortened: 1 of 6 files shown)


client/flutter/lib/pages/travel_advice.dart
```diff
@@ -1,12 +1,11 @@
+import 'package:font_awesome_flutter/font_awesome_flutter.dart';
```
hspinks, line 1: Are we using this instead of the built-in Material icons because we're moving away from Material? Otherwise, we can use the built-in `check_circle` and `warning` icons

client/flutter/lib/pages/travel_advice.dart
```diff
@@ -56,55 +55,50 @@ class _TravelAdviceState extends State<TravelAdvice> {
...
-          child: Text(
-            S.of(context).travelAdvicePageListTitle,
-            style: TextStyle(
-                fontSize: 24,
-                fontWeight: FontWeight.w800,
-                color: Color(0xff050C1D)),
-          ),
-        ),
-      ),
+          padding: const EdgeInsets.all(26.0),
```
hspinks, line 58: Another super-nit: gutters/paddings should be in multiples of 4. Definitely not a blocker here, though

advayDev1: Changed to 28.  However David's mock uses 26, 29, and more.

client/flutter/lib/pages/travel_advice.dart
advayDev1: Cupertino doesn't even have a warninng icon sadly.
FontAwesome looks far better than Material:
FontAwesome:
Screen Shot 2020-04-26 at 2 34 22 PM



Material:

Screen Shot 2020-04-26 at 2 38 00 PM

advayDev1:
ptal @hspinks

PR #1134, opened 2020-04-26
Remove references to material

theswerd:
### Remove any non-essential references to material

As we are using a CupertinoApp rather than a Material App, and our app is an iOS centric design, I replaced any reference to Material with Cupertino that didn't have to be Material.

Places were material dependencies are still within the app:
- dialogs.dart: As we don't have designs for the dialogs they are unchanged
- news_feed_item: As NewsFeedItem has an `InkWell` it needs Material
- swipeable_open_container.dart: Uses Material `ShapeBorderTween`
- carousel_slide.dart: Uses the Material `lightbulb_outline` icon
- about_page.dart: Uses the Material `showLicensePage` method
- question_index.dart: Uses Material `ExpansionTile` and Material  `add_circle_outline` icon
  
#### Screenshots
These are screenshots of different parts of the app, they are all fully functional even with the change to Cupertino.
Simulator Screen Shot - iPhone 11 Pro Max - 2020-04-25 at 18 05 19
Simulator Screen Shot - iPhone 11 Pro Max - 2020-04-25 at 18 05 35
Simulator Screen Shot - iPhone 11 Pro Max - 2020-04-25 at 18 06 23

* [x] iOS Simulator
* [ ] Android Emulator
* [ ] iOS Device
* [ ] Android Device


advayDev1:
@theswerd you should run `./run-dev-setup.sh` from the root of the repository.  It will prevent you from sending any PRs that have broken tests/formatting/etc.

Issue #1130, opened 2020-04-25
UI Refresh: Update all page headers to match spec


advayDev1:
This item will track fixing all the page headers.

advayDev1:
@davidkaneda can we get Ben a spec on exactly how our page headers should work, esp as they get scrolled off the screen

davidkaneda:
Closing for now, used Material: #1277

PR #1137, opened 2020-04-27
Partial fix headers #1130; plug in temp content

advayDev1:
Mocks use 36, 40, and 48 font, and they are not final.  I picked the one that works the best for now.  Also travel advice can now match the mock re: header color.\

Temp content makes the app feel functional again

ImprovedHeader

Screen Shot 2020-04-26 at 5 40 18 PM

### DIFF
--- a/client/flutter/lib/pages/main_pages/app_tab_router.dart
+++ b/client/flutter/lib/pages/main_pages/app_tab_router.dart
@@ -1,12 +1,14 @@
 import 'package:firebase_analytics/firebase_analytics.dart';
 import 'package:flutter/cupertino.dart';
 import 'package:flutter/material.dart';
+import 'package:who_app/api/content/schema/advice_content.dart';
+import 'package:who_app/api/content/schema/fact_content.dart';
 import 'package:who_app/constants.dart';
 import 'package:who_app/pages/latest_numbers.dart';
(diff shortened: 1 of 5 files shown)


client/flutter/lib/pages/travel_advice.dart
```diff
@@ -31,6 +31,9 @@ class _TravelAdviceState extends State<TravelAdvice> {
       return;
     }
     Locale locale = Localizations.localeOf(context);
+    print(context);
+    print(locale);
+    print(widget);
```
hspinks, line 36: remove?

```suggestion

```

client/flutter/lib/pages/main_pages/app_tab_router.dart
```diff
@@ -27,9 +29,9 @@ class AppTabRouter extends StatelessWidget {
           case 1:
             return LatestNumbersPage();
           case 2:
-            return LearnPage();
+            return ProtectYourself(dataSource: FactContent.protectYourself);
           case 3:
-            return CheckUpPage();
+            return TravelAdvice(dataSource: AdviceContent.travelAdvice);
```
hspinks, line 34: Do we need to do this? I think it's ok to leave the placeholders, we'll be filling them in shortly anyway...

PR #1146, opened 2020-04-27
First pass at symptom checker.

patniemeyer:
This PR introduces a data model and view model for the symptom checker feature.  The views are functional but no time was spent styling these so please do not review them for stylistic concerns.

This PR includes YAML-driven, localizable, question and answer views supporting (currently) four types of questions: a) Yes/No, b) Short list single selection, c) Short list multiple selection, d) Long list single selection.  Adding additional question types is intended to be easy.  A simple example of how the questions results may be analyzed is also included.

A separate PR will introduce the `display_condition` expressions that will allow individual questions and answers to be dynamically gated.

A separate PR will implement the static API generation tool that was proposed to bind the YAML question and answer ids to the summary logic.


client/flutter/assets/content_bundles/symptom_checker.en.yaml
```diff
@@ -0,0 +1,21 @@
+schema_version: 1
+content_version: 1
+contents:
+  type: symptom_checker
+  items:
+    -
+      id:  question_1
```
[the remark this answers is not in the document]
patniemeyer: @brunobowden Not sure you got a chance to look at the full proposal (the PR just has a minimal test) but I think we can cover what you described above easily with it:  https://docs.google.com/document/d/1c7IvXZ2JNCt2vb1cVr5M2Gx5V8r3cU2E-PplTcsPxtY/edit#heading=h.b1psbq4oi4iu

We discussed the possibility of implementing the full SurveyJS format but it seems like it is way more complicated than what we need. We also talked about supporting a subset of SurveyJS, but that seems like it would be the worst of both worlds - incompatibility without the flexibility to tailor for things we use.

client/flutter/assets/content_bundles/symptom_checker.en.yaml
```diff
@@ -0,0 +1,21 @@
...
+      type: yes_no
```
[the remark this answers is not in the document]
patniemeyer: @brunobowden My thinking was that people are used to seeing "yes/no" options as side-by-side buttons.  The way we have the API structured in this PR it is just a few lines of code to add a new question type, so I went ahead and used the simplest one possible for the example.  But you are right that the short list format also covers this and if we don't want it we can easily remove it.

advayDev1:
@brunobowden: @hspinks and i reviewed and discussed Pat's proposal posted in Slack symptom-checker.  This is the design we are going with and will not be using surveyjs or tencent's.

client/flutter/lib/pages/symptom_checker/question_pages/short_list_question_view.dart
```diff
@@ -0,0 +1,129 @@
...
+            Material(
+              color: Colors.white,
+              child: _allowsMultipleSelection
+                  ? Checkbox(
+                      value: _multipleSelections.contains(answer.id),
+                      onChanged: (_) {},
+                    )
+                  : Radio<String>(
+                      groupValue:
+                          _singleSelection == answer.id ? answer.id : "",
```
hspinks, line 90: Again I think just `_singleSelection` should suffice

```suggestion
                      groupValue:  _singleSelection,
```

client/flutter/lib/pages/symptom_checker/question_pages/short_list_question_view.dart
```diff
@@ -0,0 +1,129 @@
...
+                      value: answer.id,
+                      onChanged: _selected),
```
hspinks, line 92: Is this needed? To stay consistent with the other elements, suggest:

```suggestion
                      onChanged: (_) {},),
```

client/flutter/lib/pages/symptom_checker/symptom_checker_view.dart
```diff
@@ -0,0 +1,157 @@
...
+  @override
+  void answerQuestion(Set<String> answerIds) {
+    _model.answerQuestion(answerIds);
+  }
+
+  /// Receive back indication from the page and update the model.
+  @override
+  void goBack() async {
+    await _previousPage();
+    _model.previousQuestion();
```
hspinks, line 155: Since the model's `previousQuestion` will notify this class as a listener to call `_nextPage()`, could we simplify this flow and remove the need for the `_previousPage()` call? It feels more unidirectional if the PageController is only ever an expression of the model's last page index. (Not sure if I'm expressing myself clearly here, let me know if this makes sense)

client/flutter/lib/pages/symptom_checker/question_pages/long_list_question_view.dart
```diff
@@ -0,0 +1,105 @@
+import 'package:flutter/cupertino.dart';
+import 'package:flutter/material.dart';
+import 'package:flutter_html/flutter_html.dart';
+import 'package:who_app/api/content/schema/symptom_checker_content.dart';
+import 'package:who_app/pages/symptom_checker/question_pages/previous_next_buttons.dart';
+import 'package:who_app/pages/symptom_checker/symptom_checker_model.dart';
+
+// A long list question supporting a single selection.
+class LongListQuestionView extends StatefulWidget {
```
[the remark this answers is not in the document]
patniemeyer: Definitely!  Either a superclass or a mixin.  I held off on doing this in this PR simply to avoid constraining the direction that the views would take and to try to make the PR one notch less abstract for the review...  but yep, that would be at the top of my list for the "real" views.

client/flutter/lib/pages/symptom_checker/symptom_checker_view.dart
patniemeyer: This is a good question.  So, the reason I kept the view in the loop here was two-fold:  1) So that the UI can perform arbitrary animated transitions to move to the next/previous page.  2) I wanted to preserve the option for us to allow the user to simply drag the page back to go backwards in the page view rather than having to use the button.  In fact if you simply remove the `NeverScrollableScrollPhysics` it will allow you to do this.  I just did not hook up logic to notice that and inform the model to revert a page.

TLDR: I was really unclear on what kind of navigation the designers are going to want for this and I wanted to keep it flexible.  

A related topic - If you go back and then immediately go forward again should it preserve your previous choice on the second page or clear it?  Right now it is clearing it, but technically it doesn't have to unless choice of question changes.

client/flutter/lib/pages/symptom_checker/question_pages/long_list_question_view.dart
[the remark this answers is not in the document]
hspinks: Makes sense! Marking as resolved 👍

client/flutter/lib/pages/symptom_checker/symptom_checker_view.dart
hspinks: Makes a lot of sense - to the first point, could we handle arbitrary transitions in the `_nextPage` call? But, I think it makes sense to keep it the way you've implemented for now, and we can always re-examine when we know more about the transitions.

@davidkaneda & @asianfoodlab - see Pat's question above about whether going back and then forward should preserve answer choices, would love to get your thoughts! One thing to keep in mind is that in the future, when/if we have conditional logic on whether or not to show questions, we may not even show the same question if the user goes back and changes their answer.

davidkaneda: Would bank on saving the current answers if the user goes back

client/flutter/assets/content_bundles/symptom_checker.en.yaml
```diff
@@ -0,0 +1,95 @@
...
+      question_html: |
+        <p>Yes - No auestion 1</p>
```
advayDev1, line 10:
```suggestion
        <p>Yes - No question 1</p>
```

client/flutter/lib/api/content/schema/symptom_checker_content.dart
```diff
@@ -0,0 +1,138 @@
...
+  SymptomCheckerQuestion(
+      {@required this.type,
+      @required this.id,
+      @required this.questionHtml,
+      @required this.imageName,
+      @required this.displayCondition,
+      this.answers});
+}
+
+class SymptomCheckerAnswer {
```
advayDev1, line 126: Please add an icon attribute

client/flutter/lib/api/content/schema/symptom_checker_content.dart
```diff
@@ -0,0 +1,138 @@
...
+  /// An id for the question that is unique within its question context.
+  final String id;
+
+  /// An expression indicating whether the question should be presented.
+  final String displayCondition;
+
+  /// The answer
+  final String answerHtml;
```
advayDev1, line 134: Since this is in the answer class, think it should be bodyHtml

client/flutter/lib/api/content/schema/symptom_checker_content.dart
```diff
@@ -0,0 +1,138 @@
...
+    }
+    throw Exception("should be unreachable");
+  }
+
+  SymptomCheckerQuestion(
+      {@required this.type,
+      @required this.id,
+      @required this.questionHtml,
+      @required this.imageName,
+      @required this.displayCondition,
```
advayDev1, line 122: Please assert that this is null or empty given that the evaluator is not implemented.

client/flutter/lib/pages/symptom_checker/symptom_checker_view.dart
patniemeyer: @hspinks re: `_nextPage` So, the thing that is currently muddying this is that the model has any notion of the "current page"...  We can change the API to let the view update any page of the model with new answers and have it simply do the right thing (e.g. change or truncate subsequent answers).  However the view will still need to participate in the user's responses in order to maintain its own notion of the current page.  e.g. take the case where the user goes back a page and then goes forward again without making changes... In that case there is no change to the model, only the view's notion of the current page.  So the view will always need to be in the loop and this makes sense if you think of it as a "controller" (in the classic MVC sense).

client/flutter/lib/api/content/schema/symptom_checker_content.dart
patniemeyer: Added with an example in the short list views.
image

client/flutter/lib/api/content/schema/symptom_checker_content.dart
patniemeyer: Done.

client/flutter/lib/api/content/schema/symptom_checker_content.dart
patniemeyer: Done.

client/flutter/lib/pages/symptom_checker/question_pages/short_list_question_view.dart
```diff
@@ -1,6 +1,7 @@
 import 'package:flutter/cupertino.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter_html/flutter_html.dart';
+import 'package:flutter_svg/svg.dart';
```
hspinks, line 4: super-nit / not a blocker: other files import from `flutter_svg/flutter_svg.dart` which [looks like](https://github.com/dnfield/flutter_svg/blob/master/lib/flutter_svg.dart) it just exports the `svg.dart` file, so you could import from the same:

```suggestion
import 'package:flutter_svg/flutter_svg.dart';
```

client/flutter/lib/pages/symptom_checker/symptom_checker_view.dart
hspinks: Makes sense to me. I think that if we do want to update this to make sure the model is the source of truth of the current page, then that could happen in a different PR when/if we feel like we have a need for that

client/flutter/lib/pages/symptom_checker/question_pages/short_list_question_view.dart
advayDev1: @hspinks don't think we should do it in this PR, due to:

Screen Shot 2020-05-05 at 11 04 08 PM

PR #1151, opened 2020-04-28
[feat] Create ThemedText widget for default text styles

### LINKED ISSUE #1135: UI Refresh: Consistently use a theme
Our code will use a defined consistent style for as many components as possible, but most importantly for text (headings of each level, body text, button txt, etc.)

@davidkaneda it appears your most recent prototype figmas do have a heading-level-based style guide, and we'd like to make sure all text is one of those styles, or a customization of those styles.  Our request is that when you finalize the mocks all elements that can, use one of your defined them styles and you just let us know in the CSS comment which it is.

hspinks:
Creates a `ThemedText` widget wrapping the `Text` widget with default styles for different variants. Variants use `h1`, `h2`, `body`, etc terminology since that is how the styles are defined in the design spec.

Note that I decided to create this new widget rather than using the [`CupertinoTextThemeData` class](https://api.flutter.dev/flutter/cupertino/CupertinoTextThemeData-class.html) because the styles there didn't match up with the design spec. I did set the theme's `bodyText` though.

Todo as part of this PR:
- [x] Create an `AutoSizeThemedText` widget to do the same as `ThemedText` using an `AutoSizeText` internal widget instead of `Text`
- [x] Update color constants
- [ ] ~Possibly create a standardized way of handling gutter spacing / padding?~

Closes #1135 

---

- [x] REQUIRED: Do you have an Issue **assigned to you** for this PR?


client/flutter/lib/components/themed_text.dart
```diff
@@ -0,0 +1,187 @@
...
+    this.softWrap,
+    this.strutStyle,
+    this.style,
+    this.textAlign,
+    this.textDirection,
+    this.textScaleFactor,
+    this.textWidthBasis,
+  }) : super(key: key);
+
+  static final Map<TypographyVariant, TextStyle> variantStyles = {
```
britannio, line 40: If we use a function containing a switch statement with all the `TypographyVariant ` values then as long as it doesn't contain the `default` clause, a warning is created. A map doesn't enforce all enum values to be specified.

hspinks: Good point. I'll use a function & switch instead

client/flutter/lib/components/themed_text.dart
```diff
@@ -0,0 +1,209 @@
...
+  final TextStyle style;
+  final TextAlign textAlign;
+  final TextDirection textDirection;
+  final double textScaleFactor;
+  final TextWidthBasis textWidthBasis;
+
+  ThemedText(
+    this.data, {
+    Key key,
+    @required this.variant,
```
advayDev1, line 26: nit move required to front

client/flutter/lib/components/themed_text.dart
```diff
@@ -0,0 +1,209 @@
...
+  final TextAlign textAlign;
+  final TextDirection textDirection;
+  final Key textKey;
+  final double textScaleFactor;
+  final bool wrapWords;
+
+  AutoSizeThemedText(
+    this.data, {
+    Key key,
+    @required this.variant,
```
advayDev1, line 162: same nit

PR #1163, opened 2020-04-30
Prototype of Symptom *tracker*

advayDev1:


advayDev1:
@hspinks @asianfoodlab As far as I could get.  And now for my 1.5 hrs of sleep 🤗 
sym-tracker-PROTOTYPE

asianfoodlab:
Amazing work @advayDev1 , @hspinks.  Thank you!

client/flutter/pubspec.yaml
```diff
@@ -49,6 +49,7 @@ dependencies:
   path_provider: ^1.6.5
   fl_chart: ^0.8.7
   s2geometry: 0.1.1
+  flutter_secure_storage: 3.3.3
```
britannio, line 52: https://github.com/mogol/flutter_secure_storage/issues/69 
This package has several cryptography issues that have been ignored since July/August 2019 and apparently it crashes on some devices
https://github.com/mogol/flutter_secure_storage/issues

advayDev1: Good to know!

client/flutter/lib/components/page_button.dart
```diff
@@ -32,8 +33,12 @@ class PageButton extends StatelessWidget {
 
   @override
   Widget build(BuildContext context) {
-    return CupertinoButton(
```
theswerd, line 35: Is there a reason this was changed here?

client/flutter/pubspec.yaml
britannio: Maybe it's worth forking and cleaning up the package then auditing it with some security folks?

client/flutter/lib/pages/main_pages/check_up_page.dart
```diff
@@ -1,14 +1,342 @@
...
+class CheckUpPage extends StatefulWidget {
+  @override
+  _CheckUpPageState createState() => _CheckUpPageState();
+}
+
+class _CheckUpPageState extends State<CheckUpPage> {
+  List<SymptomResult> timeseries;
+
+  _CheckUpPageState() : super() {
+    SymptomTimeseries().addListener(refreshDashboard);
```
britannio, line 22: IMO it's more intuitive to update data within the `ChangeNotifier`, call `notifyListeners()` then access `FooChangeNotifier.data` via `.addListener` and call setState. 

A read shouldn't be necessary after every write.

Or even better, https://www.youtube.com/watch?v=s-ZG-jS5QHQ, use ValueListenableBuilder
docs: https://api.flutter.dev/flutter/widgets/ValueListenableBuilder-class.html

advayDev1:
The PR rises, like a zombie ;)

Nathan-55:
can the app prescribe treatment to the patients in case of any detected symptom?

PR #1181, opened 2020-05-04
Add app_en for Exposure Notifications app, and exposure_notifications for plugin

advayDev1:
Add app_en for Exposure Notifications app, and exposure_notifications for plugin

Closes WorldHealthOrganization#1083
Affects WorldHealthOrganization#1167

Screen Shot 2020-05-03 at 10 41 16 PM

Screen Shot 2020-05-03 at 9 18 46 PM


.github/workflows/client-build-en.yaml
```diff
@@ -0,0 +1,47 @@
+name: Client Build with EN
+on:
+  push:
+    paths:
+      - '.github/workflows/**'
+      - 'api/**'
+      - 'client/exposure_notifications/**'
+      - 'client/app/**'
+      - 'client/app_en/**'
+      - 'client/flutter/**'
```
hspinks, line 10: I'm not sure we want to run this CI for changes to the non-EN-flavored app... building the client once already takes 8-10 minutes or more, this will slow down our CI substantially

.github/workflows/client-test-en.yaml
```diff
@@ -0,0 +1,56 @@
+name: Client Tests with EN
+on:
+  push:
+    paths:
+      - '.github/workflows/**'
+      - 'api/**'
+      - 'client/exposure_notifications/**'
+      - 'client/app/**'
+      - 'client/app_en/**'
+      - 'client/flutter/**'
```
hspinks, line 10: Same comment on tests... these are faster at ~2 min/run, but every min counts in our CI...

client/app
```diff
@@ -0,0 +1 @@
+./flutter
```
hspinks, line 1: I know we're trying to keep things consistent with `app` and `app-en`, but it feels a little weird to have a symlink here... maybe change `app-en` to `flutter-en`? Or, we (eventually) just move the entire `flutter` dir to `app` (coordinated in advance to minimize the open PRs that will have issues with the files moving)

client/.gitignore
```diff
@@ -28,11 +28,11 @@
 .packages
 .pub-cache/
 .pub/
-/build/
+build/
```
hspinks, line 31: since we are duplicating the platform-specific gitignores, does it make more sense to just duplicate this gitignore in the `app-en` dir too? alternatively, does git allow symlinking to a gitignore?

.github/workflows/client-build-en.yaml
advayDev1: Let me run a test.  A primary purpose for this change is that neither v1 nor EN break each other.  Let me see what the total difference in time is.

.github/workflows/client-test-en.yaml
advayDev1: Because the tests run in parallel with the other tests , and because tests take strictly less time than the build it appears to cause no added wait time.  See:
Screen Shot 2020-05-04 at 9 29 48 AM

.github/workflows/client-build-en.yaml
advayDev1: They do run in parallel but it does affect our github repo budget.  Will remove the app builds for now.

client/app
advayDev1: Using the symlink means we can do the right thing and call it app and app_en everywhere (workflows, the pubspec paths, etc.), except purely in the folder.

I've filed https://github.com/WorldHealthOrganization/app/issues/1183 to remove flutter once and for all.

client/.gitignore
advayDev1: Yeah I'll just dupe them...

.github/workflows/client-build-en.yaml
advayDev1: What I did is make the build run only on a pubspec change.

advayDev1:
ptal @hspinks

Issue #1106, opened 2020-04-23
Symptom Checker: Create simple “risk assessment” results engine


advayDev1:
**The product definition, design, content, and assets are a WIP and are neither finalized nor fully approved by WHO. The purpose of this task is to prototype the feature so we can execute quickly when the non-engineering deliverables are finalized.**

advayDev1:
This engine shall be hardcoded for v1.  When WHO provides the algorithm it will be updated in code.

PR #1206, opened 2020-05-08
Symptom Checker Logic - v5

### LINKED ISSUE #1105: Symptom Checker: Conditionally navigate between survey screens using defined logic
Hard-coded logic.  Single option for all countries

advayDev1:
Closes #1105
Affects #1106 

Removes question types not intended to be designed in launch v1.
Allow no selection for multiple selections (none of the above option added in separate UI pr).

Implements PM spec of symptom checker v5 logic for questions.

I used a small subset of JS/Dart expressions in https://pub.dev/packages/expressions which is super short and does everything we need.

Will write the results selector based on the same expressions.

Most important thing to verify is that the logic matches:
https://docs.google.com/spreadsheets/d/1JSkO2Hg-DqnaVdoj3HtBxip-zzuQy4EwcZV6tYHMCNA/edit#gid=1910414175

TESTED = Android simulator for each path through the above spreadsheet

What's nice about this is we can add in the user's country or other facets inside if we want to.

Our DSL is:
```
/// Condition expressions may use bool and int literals, and the following
/// names (where questionId is the id of any prior question and
/// answerId is any answer id in that question):
/// - wasDisplayed("<questionId>") - whether a prior question was displayed
///
/// The following are defined solely if the question was displayed
/// (conditions must not rely on answers to questions that may not have been
/// displayed):
/// - int : selectionCount("<questionId>") - the number of answers selected for
///       the given question (0 for none of the above)
/// - bool : <questionId>["<answerId>"] - whether or not the given answer
///       was selected.
///
/// Conditions may use any of the following operators:
/// - bool || && !
/// - bitwise | & ^ << >> ~
/// - equality == != <= >= < >
/// - arithmetic + - * / %
```


advayDev1:
Now v5 :) ptal @hspinks

advayDev1:
> Main question is what happens with invalid expressions? E.g. non-existent function or invalid id for question or answers.
> 
> It should probably cancel the symptom checker since it's no longer operating correctly. Reporting a Google Analytic error would also seem appropriate.

@brunobowden I did not record the error to GA so we don't infer where a particular user failed in the flow, but I did make it show an error and quit.  We can get PM input on the exact warning later.   (all of the styling is in a future PR)


Screen Shot 2020-05-09 at 2 45 36 PM

.vscode/settings.json
```diff
@@ -0,0 +1,6 @@
+{
+  "files.exclude": {
```
advayDev1, line 2: re: our symlink directories

client/flutter/lib/pages/symptom_checker/symptom_checker_model.dart
```diff
@@ -38,25 +37,33 @@ class SymptomCheckerModel with ChangeNotifier {
...
-    // TODO: This should check the display conditions.
-    // No more questions, series complete.
-    if (_content.questions.length == pages.length &&
-        currentPage.selectedAnswers.isNotEmpty) {
-      isComplete = true;
+    bool nextPageFound = false;
+    for (var i = pages.length;
+        !nextPageFound && i < _content.questions.length;
+        i++) {
+      nextPageFound = SymptomLogic().evaluateCondition(
```
hspinks, line 56: Nit, but do we want to store a reference to an instance SymptomLogic rather than re-creating every time? Definitely not a blocker on this PR though

client/flutter/lib/pages/symptom_checker/symptom_checker_model.dart
```diff
@@ -38,25 +37,33 @@ class SymptomCheckerModel with ChangeNotifier {
   /// the pages list. If the series is complete the isComplete flag will be set.
   /// In both cases the change notifier will fire to indicate the update.
   void answerQuestion(Set<String> answerIds) {
+    try {
+      isComplete = _answerQuestionImpl(answerIds);
```
hspinks, line 41: If there is some error in computing next question with the logic, but it isn't caught by this try/catch, we risk thinking that the user has completed the full survey when they haven't... I can't foresee exactly why that would happen, but just noting it here as a risk

client/flutter/assets/content_bundles/symptom_checker.en.yaml
```diff
@@ -4,94 +4,141 @@ contents:
...
         -
-          id: answer_2
+          id: sore_throat
+          icon_name: snowflake
           body_html: |
-            <p>Answer 2</p>
+            <p>Sore throat</p>
         -
-          id: answer_4
+          id: diarrhoea
```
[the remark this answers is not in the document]
advayDev1: The spreadsheet says Diarrhoea, and apparently that is a valid spelling https://en.wikipedia.org/wiki/Diarrhea

client/flutter/assets/content_bundles/symptom_checker.en.yaml
```diff
@@ -4,94 +4,141 @@ contents:
...
+          icon_name: snowflake
           body_html: |
-            <p>Answer 4</p>
+            <p>Diarrhoea</p>
```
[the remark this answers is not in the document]
advayDev1: The spreadsheet says Diarrhoea, and apparently that is a valid spelling https://en.wikipedia.org/wiki/Diarrhea

client/flutter/lib/pages/symptom_checker/symptom_checker_model.dart
advayDev1: Good point, done.

client/flutter/lib/pages/symptom_checker/symptom_checker_model.dart
advayDev1: not sure how that's possible with Dart's spec, but i set isFatalError to true before the try, and set it to false after isComplete to be absolutely sure.

PR #1234, opened 2020-05-11
Update flutter to 1.17.0 stable

advayDev1:


advayDev1:
@hspinks think over what we discussed and merge if you want :)

Issue #1190, opened 2020-05-05
Remove location/GPS permissions from client


advayDev1:
Regardless of the outcome of #1187 , location permissions need to be removed

Issue #1187, opened 2020-05-05
v1 Location Product Spec


advayDev1:
We need to understand exactly what the product requirements for location in v1 are, including:
* push notifications
* jurisdiction-specific policies/content
* jurisdiction-specific branding

We will not use GPS/location permissions anymore.

Here are some tech options: https://docs.google.com/document/d/1QEXeA4fgNFD_xVXljXO3tX61EEW8adwYjHCpiIMbluU/edit#

advayDev1:
@asianfoodlab @hspinks from decision log

New Decision: We will not automatically detect user’s location; we will mandate a country selection during onboarding, and nothing more.

Prior plan:
• User opts-in to share their location during onboarding.  Behind the scenes, WHO App sends rough (to preserve privacy) GPS coordinates to server

New information that required us to reevaluate and come to this new decision:
• If we want to use the Apple & Google exposure notification API in the future, it legally prohibits WHO App from detecting user's location this way.  Although we have not decided 100% that we will use Apple and Google's APIs, we do not want to preclude ourselves from using it in the future
• Suggestion from Kore that getting member states on board with launch will be easier if they have some measure of custom control over content / the ability to target country-specific content

Discussion:

v1 Use Case for User's Location:
• Nigeria has new content specific to their country
• Nigeria provides the content to WHO app team, and it gets updated in code
• A push notification can be sent to all users that have set their country as Nigeria to inform them that new content is available.
• When city-specific content eventually comes on board, it is likely to come online country by country and we can ask for a user’s city when the relevant feature is implemented, later (aligns with Kore recommendation).

Requirements:
Onboarding: UI for user to select their country (required)
• Don’t ask them for their city
• Use device language-locale to pre-populate this field based on language-locale

Settings: UI to allow users to change their country

PR #1255, opened 2020-05-12
[feat] Country select in onboarding

### LINKED ISSUE #1194: New location (country selector) onboarding screen
To sync with @davidkaneda 1130 PDT May 7 2020

hspinks:
User can select their country during the onboarding flow rather than requesting GPS location permission. Uses the [ISO 3166 alpha-2 country codes](https://www.iso.org/iso-3166-country-codes.html), and defaults to the user's device locale. Implemented to the spec defined in #1187. 

Closes #1194

There are still a number of open tasks to be done after this PR is merged:

* [ ] Communicate the user-selected country to the back end for push notifications
* [ ] Use the user-selected country for dynamic content yaml bundle (#1217) 
* [ ] Remove location permissions from the client (#1190)
* [ ] Clean up localization by extracting hard-coded strings and updating keys in the ARB for the localized strings used to be more relevant
* [ ] Localize new YAML file, load list of countries from localized file
* [ ] Decide whether we will implement a feature in v1 for a user to be able to change their country selection later after they have already been through onboarding
* [ ] Better handle errors if something goes wrong reading the YAML country list file

#### Screenshots
<details>
<summary>Country Select Page</summary>

Simulator Screen Shot - iPhone 11 Pro Max - 2020-05-12 at 02 10 46
</details>

<details>
<summary>Country List Page</summary>

Simulator Screen Shot - iPhone 11 Pro Max - 2020-05-12 at 02 11 26
</details>

* [x] iOS Simulator

---

- [x] REQUIRED: Do you have an Issue **assigned to you** for this PR?


hspinks:
FYI this PR is blocked because it includes an icon that we are awaiting legal feedback on (cc @advayDev1)

client/flutter/lib/pages/onboarding/onboarding_page.dart
```diff
@@ -19,12 +21,33 @@ class _OnboardingPageState extends State<OnboardingPage> {
...
+  void initState() {
+    super.initState();
+    setupCountries();
+  }
+
+  void setupCountries() async {
+    _couldLoadCountries = await IsoCountryList().loadCountries();
+    if (_couldLoadCountries) {
+      _countries = IsoCountryList().countries;
+      final currentCountryCode = Localizations.localeOf(context).countryCode;
```
hspinks, line 45: It feels strange accessing the context from initState, but this works and [Flutter docs](https://api.flutter.dev/flutter/widgets/State/context.html) say context is created before initState is called, but I also see [SO answers](https://stackoverflow.com/a/49458289/2434476) [disagreeing](https://stackoverflow.com/questions/51766977/flutter-showdialogcontext-in-initstate-method#comment90496359_51766977) on this. Also, I wonder if it could cause problems to cause `setState` before the widget renders, which is possible here. This is working so far, but curious if @patniemeyer, @theswerd, or @britannio have any concerns or better approaches to this.

client/flutter/lib/pages/onboarding/onboarding_page.dart
```diff
@@ -46,19 +69,52 @@ class _OnboardingPageState extends State<OnboardingPage> {
...
+          NotificationsPage(onNext: _onDone),
         ],
       ),
     );
   }
 
+  void _selectCountry(IsoCountry country) async {
+    _selectedCountry = country;
+    await setState(() {});
+    await _toPreviousPage();
```
hspinks, line 93: @davidkaneda I am automatically sending users back to the previous page as soon as they choose a country from the list. Also, the page transition looks like a "back" transition since technically we are navigating back a page. Let me know if those two behaviors are what you want here!

client/flutter/lib/pages/onboarding/onboarding_page.dart
britannio: instead of calling it directly in `initState` you can wrap it in `WidgetsBinding.instance.addPostFrameCallback((_) {...})` and put that in initState. It'll be called once after the first frame.

Also, since that's asynchronous I'd make sure that `mounted` is true as the widget could be disposed of before the asynchronous code completes.

client/flutter/assets/onboarding/iso_countries.en.yaml
```diff
@@ -0,0 +1,499 @@
+countries:
+  - name: Afghanistan
```
advayDev1, line 2: easiest to sort by the key (the code) instead of th localizabl name

client/flutter/assets/svg/streamline-onboarding-location.svg
```diff
@@ -0,0 +1,10 @@
+<svg width="145" height="145" viewBox="0 0 145 145" fill="none" xmlns="http://www.w3.org/2000/svg">
```
advayDev1, line 1: @davidkaneda 's planning on using flat  style for onboarding anyway right?
might as well get this submitted without streamline yes?

client/flutter/lib/api/iso_country.dart
```diff
@@ -0,0 +1,56 @@
...
+
+  static final IsoCountryList _singleton = IsoCountryList._internal();
+
+  factory IsoCountryList() {
+    return _singleton;
+  }
+
+  IsoCountryList._internal();
+
+  static Future<Map<String, IsoCountry>> countriesFromYaml(
```
advayDev1, line 19:
```suggestion
  static Future<Map<String, IsoCountry>> _countriesFromYaml(
```

client/flutter/lib/api/iso_country.dart
```diff
@@ -0,0 +1,56 @@
...
+    });
+    return countries;
+  }
+
+  Future<bool> loadCountries() async {
+    if (countries != null && countries.isNotEmpty) {
+      return true;
+    }
+    try {
+      countries = await countriesFromYaml(IsoCountryList.isoFilePath);
```
advayDev1, line 39:
```suggestion
      countries = await _countriesFromYaml(IsoCountryList.isoFilePath);
```

client/flutter/lib/api/iso_country.dart
```diff
@@ -0,0 +1,56 @@
...
+        name: countryYaml['name'],
+        alpha2Code: countryYaml['alpha_2_code'],
+      );
+      countries[country.alpha2Code] = country;
+    });
+    return countries;
+  }
+
+  Future<bool> loadCountries() async {
+    if (countries != null && countries.isNotEmpty) {
```
advayDev1, line 35: is it possible to have a race cond here

hspinks: 2 things to address this:
* Since `countries` is only really intended to be set once, I can make it a private variable with a read-only getter so that nothing else alters it
* I don't think I need the `isNotEmpty` here... I was thinking about it as a potential way to rebound from an error by re-fetching the list, since if there are no countries, the user can't get past onboarding, but that's not the best way to handle. I'll remove that call

britannio: Side note, shorthand syntax would be `countries?.isNotEmpty ?? false`

client/flutter/lib/pages/onboarding/onboarding_page.dart
davidkaneda: Both are great, thanks for the heads up!

client/flutter/assets/svg/streamline-onboarding-location.svg
davidkaneda: Yeah, that'd be fine

client/flutter/assets/onboarding/iso_countries.en.yaml
hspinks: I didn't do any sorting on this at all, I just kept it in exactly the order it was listed on the ISO site. Would you like me to sort?

client/flutter/lib/pages/onboarding/onboarding_page.dart
hspinks: @britannio updated if you want to give it another look

client/flutter/lib/api/iso_country.dart
hspinks: Updated

client/flutter/lib/api/iso_country.dart
hspinks: Updated

client/flutter/lib/api/iso_country.dart
hspinks: Updated and applied the shorthand (did not remove the `isNotEmpty` for now)

client/flutter/lib/pages/onboarding/onboarding_page.dart
britannio: perfect

client/flutter/assets/svg/streamline-onboarding-location.svg
hspinks: Removed file

PR #1280, opened 2020-05-13
[feat] Send isoCountryCode to location in putLocation request

### LINKED ISSUE #1190: Remove location/GPS permissions from client

hspinks:
Connects new country selector to back end. Sends location as soon as user selects it, which is before the push notification permission request. Awaits #1023 

Currently, does not do anything with the error other than logging if the request fails. This is because the app can still function with the country code saved to the device, so I don't think allowing a more severe failure makes sense here, but we will want to do something different in the future with handling errors most likely.

Closes #1190 

* [x] iOS Simulator - currently does nothing other than logging a 500 error until #1023 is merged

---

- [x] ~REQUIRED: Do you have an Issue **assigned to you** for this PR?~


client/flutter/lib/pages/onboarding/location_sharing_page.dart
```diff
@@ -1,81 +0,0 @@
-import 'package:who_app/api/who_service.dart';
-import 'package:who_app/generated/l10n.dart';
-import 'package:who_app/pages/onboarding/permission_request_page.dart';
-import 'package:flutter/cupertino.dart';
-import 'package:location/location.dart';
-import 'package:s2geometry/s2geometry.dart';
```
advayDev1, line 6: Remove the location permission and s2geometry deps from pubspec.yaml plz

hspinks: this is coming in a separate PR 😄

hspinks: Done

PR #1320, opened 2020-05-14
Pre-cache content bundles.

patniemeyer:
This PR introduces pre-caching of common content bundles upon acceptance of the terms of service and also upon subsequent app launches or returning to the foreground.

### DIFF
--- a/client/flutter/lib/api/content/content_loading.dart
+++ b/client/flutter/lib/api/content/content_loading.dart
@@ -1,4 +1,9 @@
 import 'dart:io';
+import 'package:who_app/api/content/schema/advice_content.dart';
+import 'package:who_app/api/content/schema/fact_content.dart';
+import 'package:who_app/api/content/schema/index_content.dart';
+import 'package:who_app/api/content/schema/question_content.dart';
+import 'package:who_app/api/content/schema/symptom_checker_content.dart';
 import 'package:who_app/api/endpoints.dart';
 import 'package:who_app/api/user_preferences.dart';
 import 'package:who_app/api/who_service.dart';
@@ -23,6 +28,18 @@ class ContentLoading {
   }
 
   ContentLoading._internal();
+
+  /// Pre-cache commonly used content bundles to ameliorate loading delay.
+  /// This method may be called at any time and will only load content when needed.
+  void preCacheContent(Locale locale) {
+    print("preCachecontent for locale: ${locale}");
+    AdviceContent.travelAdvice(locale);
+    FactContent.getTheFacts(locale);
+    IndexContent.homeIndex(locale);
+    IndexContent.learnIndex(locale);
+    QuestionContent.yourQuestionsAnswered(locale);
+    SymptomCheckerContent.load(locale);
+  }
 
   /// Load a localized content bundle loaded preferentially from the network, falling back
   /// to a local asset.  If no bundle can be found with the specified name an exception is thrown.
--- a/client/flutter/lib/api/user_preferences.dart
+++ b/client/flutter/lib/api/user_preferences.dart
@@ -28,6 +28,17 @@ class UserPreferences {
         .setBool(UserPreferenceKey.OnboardingCompleted.toString(), value);
   }
 
+  Future<bool> getTermsOfServiceCompleted() async {
+    return (await SharedPreferences.getInstance())
+        .getBool(UserPreferenceKey.TermsOfServiceCompleted.toString()) ??
+        false;
+  }
+
+  Future<bool> setTermsOfServiceCompleted(bool value) async {
+    return (await SharedPreferences.getInstance())
+        .setBool(UserPreferenceKey.TermsOfServiceCompleted.toString(), value);
+  }
+
   Future<bool> getAnalyticsEnabled() async {
     return (await SharedPreferences.getInstance())
             .getBool(UserPreferenceKey.AnalyticsEnabled.toString()) ??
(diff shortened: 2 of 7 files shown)


patniemeyer:
@advayDev1  One issue to discuss - The current implementation relies on the fact that there will be files in the language and locale combination found on the server (actual files to be cached).   In cases where we are relying on the fallback local files we will continue to try to load these each time the user hits the corresponding screen.  Question: Is the intent to stop that behavior?  Do we want to note the fact that the file did not load and stop trying to load it for a period of time?  If that is the case we can load the cache with the local asset and set the appropriate timeout.  Let me know.

patniemeyer:
I tested this by putting the files on my server and changing the base url.  The caching worked fine however this illustrated that a number of things in the app break in unpleasant ways when the app can't reach the network.  I'll file a separate issue / PR for those.

client/flutter/lib/main.dart
```diff
@@ -143,4 +148,33 @@ class _MyAppState extends State<MyApp> {
...
+      case AppLifecycleState.paused:
+      case AppLifecycleState.detached:
+        break;
+    }
+  }
+
+  /// Construct the Locale from the Intl locale string.
+  /// This allows us to get the Locale outside of the main build context.
+  Locale getLocale() {
+    var parts = Intl.getCurrentLocale().split('_');
```
epicfaace, line 171: What exactly is the difference between this and `Locale(Intl.getCurrentLocale())`?

client/flutter/lib/api/user_preferences.dart
```diff
@@ -44,6 +44,10 @@ class UserPreferences {
         .setBool(UserPreferenceKey.AnalyticsEnabled.toString(), value);
   }
 
+  Future<bool> getTermsOfServiceAccepted() async {
+    return getAnalyticsEnabled();
```
epicfaace, line 48: Is this correct? A user can accept terms of service and then disable analytics later.

client/flutter/lib/main.dart
patniemeyer: `Intl.getcurrentLocale()` returns the full locale string.  So the previous usage of `Locale(Intl.getCurrentLocale())` was incorrect as it was passing e.g. `en_US` as the language code and `null` as the country code.  This resulted in trying to load files e.g. `protect_yourself.en_US_null.yaml`.  I believe this fixes the problem and I've subbed it in for the previous usage.

client/flutter/lib/api/user_preferences.dart
patniemeyer: You are right, it's not correct. Will fix.

patniemeyer: Ok, I've added a real "terms of service" preference to track the legal page. Please note that to properly test this you'll have to reinstall the app as this flag is set only once during onboarding.

advayDev1:
@patniemeyer is #1328 good enough for you?

patniemeyer:
> @patniemeyer is #1328 good enough for you?

Looks good!  Will be easy to double check once deployed.

client/flutter/lib/api/content/content_loading.dart
```diff
@@ -24,6 +29,18 @@ class ContentLoading {
...
+  /// Pre-cache commonly used content bundles to ameliorate loading delay.
+  /// This method may be called at any time and will only load content when needed.
+  void preCacheContent(Locale locale) {
+    print("preCachecontent for locale: ${locale}");
+    AdviceContent.travelAdvice(locale);
+    FactContent.getTheFacts(locale);
+    IndexContent.homeIndex(locale);
+    IndexContent.learnIndex(locale);
+    QuestionContent.yourQuestionsAnswered(locale);
+    SymptomCheckerContent.load(locale);
```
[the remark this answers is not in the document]
patniemeyer: @brunobowden I did think about this :)  So with Dart capabilities at the moment something has to know about all of the content schema classes to either trigger the initialization, registration, or reflection on them.  (In some languages there are ways to search for annotations on classes, etc. but not in Dart yet, I believe).  

What we would have to do is have either a tiny build time script that generates a loader based on the files in the `content/schema` directory or at runtime try to infer the classes by looking the content bundle names.  I decided that a) this may be a bit of overkill at the moment since there are so few files and b) we may actually want to make a manual decision about which ones to load and when... e.g. stage them, etc. at least in the short term...  We could probably do the latter with annotations on the classes later.  Probably best to see how this goes for a bit.

client/flutter/lib/main.dart
```diff
@@ -143,4 +148,39 @@ class _MyAppState extends State<MyApp> {
       ),
     );
   }
+
+  @override
+  void didChangeAppLifecycleState(AppLifecycleState state) async {
+    switch (state) {
+      case AppLifecycleState.resumed:
+        _precacheContent();
```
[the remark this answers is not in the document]
patniemeyer: Strangely, `didChangeAppLifecycleState` does not cover the initial startup but I did cover it with a call to `precacheContent()` in the main page's init method.

client/flutter/lib/components/home_page_sections/home_page_protect_yourself.dart
```diff
@@ -7,6 +7,8 @@ import 'package:who_app/api/display_conditions.dart';
 import 'package:who_app/components/themed_text.dart';
 import 'package:who_app/constants.dart';
 
+import '../dialogs.dart';
```
hspinks, line 10: nit:

```suggestion
import 'package:who_app/components/dialogs.dart';
```

client/flutter/lib/main.dart
```diff
@@ -18,7 +18,10 @@ import 'package:who_app/pages/main_pages/routes.dart';
 import 'package:who_app/constants.dart';
 import 'package:who_app/generated/l10n.dart';
 
+import 'api/content/content_loading.dart';
```
hspinks, line 21: nit

```suggestion
import 'package:who_app/api/content/content_loading.dart';
```

PR #1325, opened 2020-05-14
🎶🦜 Server country and push notification topic registration

### LINKED ISSUE #1000: Allow Efficient targeting of notifications to whole countries

### LINKED ISSUE #1221: Server should accept only country code as location

advayDev1:
Closes #1221, closes #1000

Client already sends isoCountryCode, this saves it.
It will also register the client for topics `ALL~LOCATION` (global) and `ALL~LOCATION~isoCountryCode` (per country) topics.

In the future we could have `TESTINGCENTERS~LOCATION~isoCountryCode` etc. if users only want a few types of notifications.

[redacted] GCP traces:

```
2: {
time: "2020-05-14T01:53:50.355Z"
severity: "INFO"
logMessage: "who.NotificationsManager updateSubscriptions: SUCCESS - Subscribed d0zOdIG6[redacted]KusQJXig2uNK8gSFgxBD1hjmBGE4XyQQQvzbOVP to topic ALL~LOCATION
"
sourceLocation: {3}
}
3: {
time: "2020-05-14T01:53:50.660Z"
severity: "INFO"
logMessage: "who.NotificationsManager updateSubscriptions: SUCCESS - Subscribed d0zOdIG[redacted]KusQJXig2uNK8gSFgxBD1hjmBGE4XyQQQvzbOVP to topic ALL~LOCATION~US
"
sourceLocation: {3}
}
```

The topics are available in FCM console:

Screen Shot 2020-05-13 at 6 10 06 PM

### DIFF
--- a/server/appengine/src/main/java/who/FirebaseModule.java
+++ b/server/appengine/src/main/java/who/FirebaseModule.java
@@ -14,12 +14,11 @@ import javax.inject.Singleton;
 public class FirebaseModule extends AbstractModule {
 
   @Provides @Singleton
-  FirebaseApp provideFirebaseApp() throws IOException {
-
+  FirebaseApp provideFirebaseApp(Environment env) throws IOException {
     FirebaseOptions options = new FirebaseOptions.Builder()
       .setCredentials(GoogleCredentials.getApplicationDefault())
       // FIXME
-      .setDatabaseUrl("https://who-myhealth-staging.firebaseio.com")
+      .setDatabaseUrl("https://" + env.firebaseApplicationId() + ".firebaseio.com")
       .build();
 
     FirebaseApp.initializeApp(options);
--- a/server/appengine/src/main/java/who/FirebaseModule.java
+++ b/server/appengine/src/main/java/who/FirebaseModule.java
@@ -1,0 +1,34 @@
+package who;
+
+import com.google.auth.oauth2.GoogleCredentials;
+import com.google.firebase.FirebaseApp;
+import com.google.firebase.FirebaseOptions;
+import com.google.firebase.messaging.FirebaseMessaging;
+import com.google.inject.AbstractModule;
+import com.google.inject.Provides;
+
+import java.io.IOException;
+
+import javax.inject.Singleton;
+
+public class FirebaseModule extends AbstractModule {
+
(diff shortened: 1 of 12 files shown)


server/appengine/src/main/java/who/FirebaseModule.java
```diff
@@ -0,0 +1,33 @@
...
+
+import javax.inject.Singleton;
+
+public class FirebaseModule extends AbstractModule {
+
+  @Provides @Singleton
+  FirebaseApp provideFirebaseApp(Environment env) throws IOException {
+    FirebaseOptions options = new FirebaseOptions.Builder()
+      .setCredentials(GoogleCredentials.getApplicationDefault())
+      // FIXME
```
advayDev1, line 20:
```suggestion
```

PR #1328, opened 2020-05-14
Serve all country asset bundles from server

### LINKED ISSUE #1327: Server must serve English content bundles for all countries
cc @patniemeyer

advayDev1:
Closes #1327

http://localhost:8080/content/bundles/your_questions_answered.en_US.yaml
http://localhost:8080/content/bundles/your_questions_answered.en_FR.yaml
http://localhost:8080/content/bundles/your_questions_answered.en_KR.yaml

etc.

### DIFF
--- a/server/appengine/src/main/java/who/WhoServletModule.java
+++ b/server/appengine/src/main/java/who/WhoServletModule.java
@@ -25,6 +25,19 @@ public class WhoServletModule extends ServletModule {
     serve("/terms").with(forwardTo("/terms.pdf"));
     serve("/privacy").with(forwardTo("/privacy.pdf"));
 
+    // TODO: Make this much better.
+    final String[] bundles = new String[]{"news_index", "get_the_facts", "home_index", "learn_index", "protect_yourself", "symptom_checker", "travel_advice", "your_questions_answered"};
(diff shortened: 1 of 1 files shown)


server/appengine/src/main/java/who/WhoServletModule.java
```diff
@@ -25,6 +25,19 @@
     serve("/terms").with(forwardTo("/terms.pdf"));
     serve("/privacy").with(forwardTo("/privacy.pdf"));
 
+    // TODO: Make this much better.
+    final String[] bundles = new String[]{"news_index", "get_the_facts", "home_index", "learn_index", "protect_yourself", "symptom_checker", "travel_advice", "your_questions_answered"};
+    for (final String bundle : bundles) 
+    {
+      for (int i = 0; i < 26; i++) {
```
hspinks, line 32: This is fine for now, but thinking through how this changes in the future... do you have a plan for how this would update if a country ends up requiring a different content bundle? E.g. exclusion list of paths, reading from the actual files, etc?

advayDev1: Config file seems best, but let's discuss and design.  Just doing this to launch for now.

PR #1361, opened 2020-06-02
Isuuse #1350

### LINKED ISSUE #1350: On small screens, intro text is left aligned
What's happening here is the text widget is centered in the container, but the widget's own text alignment is left.

iPhone 8:

Simulator Screen Shot - iPhone 8 - 2020-05-23 at 17 55 02

cosmicRover:
Dynamically adjusts the font size based on screen width and platform.
Closes #1350 
```
if (Platform.isAndroid) {
      if (screenWidth > 320) {
        return 18;
      } else {
        return 16;
      }
    } else if (Platform.isIOS) {
      if (screenWidth > 375) {
        return 18;
      } else if (screenWidth <= 320) {
        return 14;
      } else {
        return 16;
      }
    }
```
#### Tested on:
* [x] iOS Simulator
* [x] Android Emulator
* [ ] iOS Device
* [ ] Android Device


cosmicRover:
need to put a default return statememt.

PR #1369, opened 2020-06-18
Create stale.yml

advayDev1:


brunobowden:
@advayDev1 - does this only mark pull requests as stale and not issues? It's not obvious from the diff and would've been helpful to make that explicit in the description. Since it's already merged, please add an explanatory comment here or in a future diff.

advayDev1:
> @advayDev1 - does this only mark pull requests as stale and not issues? It's not obvious from the diff and would've been helpful to make that explicit in the description. Since it's already merged, please add an explanatory comment here or in a future diff.

Done.

PR #1293, opened 2020-05-13
Firebase App Distribution with Fastlane

epicfaace:
## description

Setup fastlane for firebase app distribution (fixes #1044). Instructions for deploying this are in release/manual-build.md.

I've added one lane, `distribute` (for who-myhealth-production). When actually deploying to firebase app distribution, we should run `bundle exec fastlane distribute` so that the app is deployed to who-myhealth-production.

In order to test the fastlane integration, I replaced the app id in Fastfile with `1:1057717871229:android:4fbe832a3de230450dc714` (which is the app id in who-myhealth-staging) and then deployed to who-myhealth-staging.

## dependencies

Also added a Gemfile.lock -- this is needed to pin down dependencies used by fastlane (such as firebase_app_distribution).

The new dependency, `fastlane-plugin-firebase_app_distribution`, is also maintained by the fastlane team.

PR #1399, opened 2020-07-02
Push Android APK to Firebase staging on every master build

advayDev1:
Push Android APK to Firebase staging on every master build

Example success:
https://github.com/WorldHealthOrganization/app/actions/runs/155730242

Affects #1044

Unrelatedly, adds a security check to our gradle wrappers.  Read more here: https://github.com/marketplace/actions/gradle-wrapper-validation

----

### Execution

On every push to `master`, an Android APK will be pushed to the Firebase App Distribution staging environment using the `client-distribute-staging` workflow. Firebase will invite all members of the `every-build` testing group to test each of these builds.

### Setup

1.  [x] (First time) Create [a service account](https://console.cloud.google.com/iam-admin/serviceaccounts/details/104111540645440452578?project=who-myhealth-staging) on `who-myhealth-staging` GCP with `Firebase App Distribution Admin` role.
2.  [x] Create a Key for that service account: Add Key -> Create New Key -> JSON. Save the JSON file locally as `key.json`.
3.  [x] Generate a random passphrase. Store that passphrase as GitHub secret `FIREBASE_APPDEPLOY_STAGING_SVCACCT_PASSPHRASE`.
4.  [x] Encrypt and encode the key using the generated passphrase:

        gpg --symmetric --cipher-algo AES256 --armor < key.json | base64

5.  [x] Store the encrypted key as GitHub secret `FIREBASE_APPDEPLOY_STAGING_SVCACCT_JSON`.
6.  [x] Delete the older key on the service account on GCP.


.github/workflows/client-distribute-staging.yaml
```diff
@@ -0,0 +1,74 @@
+name: Firebase Staging App Distribution
+on:
+  push:
+    # FIXME: Uncomment the `branches` line before submitting.
+    # It is needed as a comment for me to test this.
+    # Master branch has the staging firebase config.
+    # branches: [ master ]
```
advayDev1, line 7: This will be uncommented PRIOR to submitting.

advayDev1:
> Don't we want to use fastlane (see #1293) instead of the firebase CLI directly?

After Fastlane is working, and we have different lanes for staging and production, yes definitely.

advayDev1:
Actually ready now, added documentation @hspinks

PR #1401, opened 2020-07-02
👊☁️ Push App Engine staging; add build info; configure environments; refactor CI

advayDev1:
Fixes #1400

Also adds build info into the app - fixes #1405

Also configures GitHub environments api - see https://github.com/WorldHealthOrganization/app/deployments and the "deployed to" entries in this PR.

Also refactors our workflows to share common elements.

-----------------

### Execution

On every push to `master` that affects the server, the server will be pushed to the staging App Engine environment using the `server-deploy-staging` workflow.

### Setup

1. [x]  (First time) Create [a service account](https://console.cloud.google.com/iam-admin/serviceaccounts/details/104111540645440452578?project=who-myhealth-staging) on `who-myhealth-staging` GCP with the following role:
    `App Engine Deployer`,
    `App Engine Service Admin`,
    `Cloud Build Editor`,
    `Cloud Scheduler Admin`, and
    `Storage Object Admin`.
2. [x]  Create a Key for that service account: Add Key -> Create New Key -> JSON. Save the JSON file locally as `key.json`.
3. [x]  Generate a random passphrase. Store that passphrase as GitHub secret `APP_ENGINE_DEPLOY_STAGING_SVCACCT_PASSPHRASE`.
4. [x]  Encrypt and encode the key using the generated passphrase:

        gpg --symmetric --cipher-algo AES256 --armor < key.json | base64

5. [x]  Store the encrypted key as GitHub secret `APP_ENGINE_DEPLOY_STAGING_SVCACCT_JSON`.
6. [x]  Delete the older key on the service account on GCP.

-------------

Build info:
Screen Shot 2020-07-08 at 6 02 43 PM

Screen Shot 2020-07-08 at 6 02 49 PM

PR #1409, opened 2020-07-16
CI: codeql-analysis.yml

advayDev1:
The workflow is mostly auto-generated by GitHub.

Also fixes the branch ref on client and server


hspinks:
Looks like the CI is failing because of `./gradlew` instead of `gradle`, might be fixed by #1401? Looks fine to me once broken CI is resolved

advayDev1:
> Looks like the CI is failing because of `./gradlew` instead of `gradle`, might be fixed by #1401? Looks fine to me once broken CI is resolved

@hspinks yep - thanks, fixed now.

PR #1449, opened 2020-08-20
📺 📚 New static content serving system

### LINKED ISSUE #1444: WhoCacheManager caches files for 7 days, when there is no cache timing header in the response
```
I/flutter (10323): Cached for https://staging.whocoronavirus.org/content/bundles/symptom_checker.en_US.yaml until 2020-08-25 23:28:31.393
I/flutter (10323): loaded https://staging.whocoronavirus.org/content/bundles/symptom_checker.en_US.yaml
```

### LINKED ISSUE #1447: Move all static content to static serving instead of appengine
- Store it in Google Cloud Storage
- Set proper cache-control headers

- When releasing to production, cut a branch and run a script.
- On every master submission, push to staging bucket via GitHub actions.
- Script will create all duplicate mappings of yaml files across countries and supported languages

- Sync it up to the cloud:
```
gsutil -m -h "Cache-Control:public, max-age=600" rsync -r ./generatedContent/ gs://who-myhealth-staging-static-content-01/content/bundles/
```

advayDev1:
Instead of using app engine, all static content will be served from Google Cloud Storage.
This lets us set the cache-control correctly.

Closes #1447 closes #1444

See https://github.com/WorldHealthOrganization/app/actions/runs/216128660 for a successful run

TESTED = I pushed updates to the text in protect yourself.  Before 10 minutes switching away from the app and switching back didn't change anything.  After 10 minutes, switching back to the app caused the new content to be visible.  Moreover, the app retained its UI state (ex. scrolling down) when the content changed.

Revised architecture:
[WHO COVID-19 System - Page 1 (1).pdf](https://github.com/WorldHealthOrganization/app/files/5100871/WHO.COVID-19.System.-.Page.1.1.pdf)


.github/workflows/static-content.yaml
```diff
@@ -0,0 +1,132 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Setup Node.js
+        uses: actions/setup-node@v1
+        with:
+          node-version: 12
+      - run: npm install
+        working-directory: tools
+      - run: DONT_PUSH=true ./_deploy-static-serving.sh
```
hspinks, line 32: out of curiosity, why the leading `_` on the script filename?

client/flutter/assets/content_bundles/get_the_facts.en.yaml
```diff
@@ -1,5 +1,5 @@
 schema_version: 1
-content_version: 1
+content_version: 3
```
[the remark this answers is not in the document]
advayDev1: i used 2 for testing, so want to be strictly higher than anything i've used.

.github/workflows/static-content.yaml
advayDev1: originally, it was a 'implementation' script, but now that i'm using it here, i'll remove the underscroe

tools/prepare-static-serving.js
```diff
@@ -0,0 +1,35 @@
...
+const files = fs.readdirSync(srcPath);
+
+const destDir = __dirname + '/staticContentBuild';
+fs.mkdirSync(destDir);
+
+for (const file of files) {
+  const regex = /^(?<base>.+)\.(?<lang>[^.]+)\.yaml$/;
+  const m = file.match(regex).groups;
+  console.log(m);
+  const expanded = countries.map((c) => `${m.base}.${m.lang}_${c}.yaml`);
```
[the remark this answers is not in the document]
advayDev1: Great point - will do.

tools/prepare-static-serving.js
```diff
@@ -0,0 +1,35 @@
...
+const files = fs.readdirSync(srcPath);
+
+const destDir = __dirname + '/staticContentBuild';
+fs.mkdirSync(destDir);
+
+for (const file of files) {
+  const regex = /^(?<base>.+)\.(?<lang>[^._]+)(?:_(?<country>[^.]+))?\.yaml$/;
+  const m = file.match(regex).groups;
+  console.log(m);
+  const expanded = m.country ? [`${m.base}.${m.lang}_${country}`] : countries.map((c) => `${m.base}.${m.lang}_${c}.yaml`);
```
hspinks, line 26: ah, one other note - do we want to do any sort of check for if the filename doesn't match, or are you thinking that throwing an error is the right thing since all files in the directory should match?

hspinks, line 26: whoops, one more note - we'll need to copy the extension without the country code as well. e.g. there should be a `protect_yourself.en.yaml`. This line is starting to get pretty dense with this suggestion so may want to expand a bit, but thinking:
```suggestion
  const expanded = m.country ? [`${m.base}.${m.lang}_${country}`] : [`${m.base}.${m.lang}.yaml`, ...countries.map((c) => `${m.base}.${m.lang}_${c}.yaml`)];
```

hspinks, line 26: Error in my earlier suggestion:
```suggestion
  const expanded = m.country ? [`${m.base}.${m.lang}_${country}.yaml`] : countries.map((c) => `${m.base}.${m.lang}_${c}.yaml`);
```

advayDev1: The plain .en. file can't be requested by any client, so we don't need this.

advayDev1: Yep, and added filtering too, PTAL

advayDev1: The js will crash for non-matchng files, which IMO it should.  It'll catch it in the CI.

hspinks: sg, just wanted to make sure it was intentional

Issue #883, opened 2020-04-05
v1.0 additional credits


brunobowden:
- [x] Searched the [existing issues](https://github.com/WorldHealthOrganization/app/issues) to ensure you are not creating a duplicate.
- [x] Followed the [Contributor Guidelines](https://github.com/WorldHealthOrganization/app/blob/master/docs/CONTRIBUTING.md).

Updated "Thanks to:" section

Credits to add:
* Biswa Gurubacharya
* Bernardo Mariano Junior
* Evan Pye
* Flavio Aggio
* Andy Pattison
* Michelle Thulkanam
* Tedros Adhanom Ghebreyesus

advayDev1:
* Michael Nichols

advayDev1:
Security team!
Will Ackerly
Jobert Abma
Michiel Prins
Peter Nancarrow
Ryan Speers
Sitora Serverova

b099l3:
This looks like it is just updating this [file](https://github.com/WorldHealthOrganization/app/blob/master/content/credits.yaml) with the names listed above.

Also there are theses todos which may include other people:
# TODO: Pre-launch, review above list to add in the full team that doesn't make git commits!
# TODO: Add businesses supporting us with donated resources.

Can I work on this as a first issue?

advayDev1:
needs to be updated with list of WHE collaborators too

tabassumwajid2005:
> others not included above or in the list already
> 
> Security:
> Will Ackerly
> Jobert Abma
> Michiel Prins
> Peter Nancarrow
> Ryan Speers
> Sitora Serverova
> Marc McLaughlin
> Andrew Nebus
> 
> PgM:
> Marie Debard
> Stephanie Liu
> 
> PM:
> Jessica Ladd
> 
> QA:
> Mohit Saini
> Tabassum Wajid

> Legal:
> Eve Chaurand
> Celine Guiillou

tabassumwajid2005:
Hi Bruno,

I have worked as a QA and contribute my efforts in this but it is the
second update of list and I am still not in the list. Is there any reason
of it or just a mistake?

Please add my name in the list of contributors and I am one of the
contributor of this effort.

Regards,
Tabassum wajid

brunobowden:
@tabassumwajid2005 - my sincere apologies. Part of the reason to put this up was to ask people to check if anyone was missing. So I'm glad you caught this. Please see this PR where in fact your name gets a very special place right next to Dr Tedros himself:

[...]

PR #1461, opened 2020-08-24
Randomize A-Z or Z-A for credits; add other leads

advayDev1:
Perfect mean equality, Cryptographically secure, even.
This way people still know where to find themselves, while preserving the notion that order != contribution.

#883

### DIFF
--- a/client/flutter/lib/pages/about_page.dart
+++ b/client/flutter/lib/pages/about_page.dart
@@ -1,4 +1,5 @@
 import 'dart:io';
+import 'dart:math';
 
 import 'package:firebase_analytics/firebase_analytics.dart';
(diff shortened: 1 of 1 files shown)


brunobowden:
@advayDev1 - this randomizes it on each release? I think that's a great way of doing this.

advayDev1:
> @advayDev1 - this randomizes it on each release? I think that's a great way of doing this.

It's different each time you click on the About the App button, so it's mean equal =D

advayDev1:
Per discussion with Bruno, leads will still be 50/50 a-z z-a, and everyone else will be a-z to be found, and a new lines has been added

PR #1469, opened 2020-08-25
Fixes #1290, Fixes #1467: Load country specific-content, and fix country list bug

### LINKED ISSUE #1290: Country dropdown doesn't work if you return to the country page for the 2nd time
**Describe the problem**
A clear and concise description of what the problem is, from a user's perspective

**To Reproduce**
1. Start the onboarding process
2. Go to the second then third page
3. Go back to the second page(via the back button)
4. Attempt to select a country
5. The country list page isn't shown

**Expected, correct behavior**
At step 4 I'd expect the page to be switched to the country list page.

**Phone/Tablet (please complete the following information):**

- Device: OnePlus 7 Pro
- OS: Android
- App Version: master

**Additional context**
This should be fixable without changing the flow but I'd prefer pushing the country list page in a new route and then using `Navigator.pop(context, <country>)` to detect which country was selected.

---

- [x] Issue has descriptive title focused on the feature or problem from the user's perspective

### LINKED ISSUE #1467: Country selector must be passed on in content locale

advayDev1:
By default we will use locale-country, but once a user selects a country,
we will actually use that country in the content YAML loading.

This starts the proper refactor of UserPreferences into a Store to use
it as an observable dependency.

Dependency: observer_producer - a verified dep I open-sourced from work

Fixes #1290, Fixes #1467

TESTED=Look at the logs as you change the country showing which YAML files are loaded.

### DIFF
--- a/client/flutter/pubspec.yaml
+++ b/client/flutter/pubspec.yaml
@@ -53,6 +53,7 @@ dependencies:
   fl_chart: ^0.8.7
   notification_permissions: ^0.4.4
   expressions: 0.1.5
+  observer_provider: 0.0.1+2
 
   intl: ^0.16.0
   flutter_localizations:
--- a/client/flutter/lib/api/content/content_loading.dart
+++ b/client/flutter/lib/api/content/content_loading.dart
@@ -1,5 +1,4 @@
 import 'dart:io';
-import 'package:who_app/api/user_preferences.dart';
 import 'package:who_app/api/who_service.dart';
 import 'package:flutter/services.dart' show rootBundle;
 import 'package:flutter/cupertino.dart';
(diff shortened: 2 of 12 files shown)


client/flutter/pubspec.yaml
```diff
@@ -53,6 +53,7 @@ dependencies:
   fl_chart: ^0.8.7
   notification_permissions: ^0.4.4
   expressions: 0.1.5
+  observer_provider: 0.0.1+2
```
britannio, line 56: Will this be documented in the future? @advayDev1

advayDev1: Haha, yes, I'll do that in 'the real world'

PR #1470, opened 2020-08-26
Code formatting for Java, Shell, JS, YAML too

advayDev1:
Prettier will format Java, Shell, JS, and YAML.  This will be done in the pre-commit hook and in the Policy CI.
Swift and Kotlin are not production ready.  Ruby is hard to identify.

If something is broken, it is most likely NOT this commit.  Look before it in the `git blame`


advayDev1:
thanks @brunobowden sadly i can't submit this due to snyk, but i'll figure that out some time...

.github/workflows/client.yaml
```diff
@@ -165,7 +165,7 @@ jobs:
       - name: Run Firebase App Distribution
         shell: bash
         run: |
-          npx firebase appdistribution:distribute --app "$FIREBASE_STAGING_APP_ID" --release-notes "$RELEASE_NOTES" --groups "every-build" app-release.apk
+          npx --no-install firebase appdistribution:distribute --app "$FIREBASE_STAGING_APP_ID" --release-notes "$RELEASE_NOTES" --groups "every-build" app-release.apk
```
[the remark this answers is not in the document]
advayDev1: Ye removed this

tools/.lintstagedrc.json
```diff
@@ -2,6 +2,15 @@
   "../**/*.md": [
     "prettier --write"
   ],
+  "../**/*.yml": [
+    "prettier --write"
+  ],
+  "../**/*.yaml": [
+    "prettier --write"
+  ],
+  "../server/appengine/src/main/java/who/**/*.java": [
```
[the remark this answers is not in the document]
advayDev1: Well it needs to avoid the vendor and built java files, which I put in the .prettierignore file, so done

advayDev1:
Ptal @brunobowden @hspinks .

advayDev1:
Also: swift is deemed not ready for production use, ruby has no extension (so its hard to find the files).  I added shell.  Kotlin too is not ready for production.

tools/.lintstagedrc.json
```diff
@@ -1,8 +1,9 @@
 {
-  "../**/*.md": [
+  
+  "../**/*.{md,java,js,sh,yaml,yml}|*.{md,java,js,sh,yaml,yml}": [
```
brunobowden, line 3: Does this work to DRY it up?
```suggestion
  "(../**/|)*.{md,java,js,sh,yaml,yml}": [
```

Infact, why doesn't the simplified `../**/*.{md,java,js,sh,yaml,yml}` work? I thought that would match on the second part of the regex for a file within the same folder.

tools/.lintstagedrc.json
```diff
@@ -1,8 +1,9 @@
...
     "prettier --write"
   ],
-  "../client/**/*.dart": [
+  "../**/*.dart|*.dart": [
```
brunobowden, line 6: Similar as above:
```suggestion
  "(../**/|)*.dart": [
```

brunobowden (approved):
@advayDev1 - this is a welcome change. Please update the description with the file formats that it covers and any other significant changes. I think it would also be good to mention why Swift and Kotlin are not includes as they aren't production ready yet.

tools/.lintstagedrc.json
advayDev1: '../**/*' doesn't work because it uses relative paths, and '../**' only prefix-matches things strictly not on this directory

i will try the parentheses approach

advayDev1: Update: parentheses does not work sadly.

Issue #1434, opened 2020-08-17
New design for Symptom poster


advayDev1:
Bruno-mocks are:
<img width="927" alt="check-up-info-only" src="https://user-images.githubusercontent.com/11729521/90442796-10345900-e090-11ea-81d2-473ca3d62534.png">

davidkaneda:
Added two options in Figma, one based on the mockup above, one as a more direct translation of the page mentioned: 
<img width="445" alt="image" src="https://user-images.githubusercontent.com/21713/92060401-86031a80-ed48-11ea-98ed-52a2c3c1cf9f.png">

PR #1476, opened 2020-08-30
For #1434 - Remove symptom checker, add symptom poster skeleton

advayDev1:
Removes the draft Symptom Checker, and adds a "poster" style design approved by WHE instead.


client/flutter/lib/api/content/content_store.dart
```diff
@@ -130,11 +130,11 @@ abstract class _ContentStore with Store implements Updateable {
...
-    final symptomChecker2 = SymptomCheckerContent(
-        await service.load(locale, countryIsoCode, 'symptom_checker'));
-    if (symptomChecker?.bundle?.contentVersion !=
-        symptomChecker2?.bundle?.contentVersion) {
-      symptomChecker = symptomChecker2;
+    final symptomPoster2 = PosterContent(
+        await service.load(locale, countryIsoCode, 'symptom_poster'));
+    if (symptomPoster?.bundle?.contentVersion !=
+        symptomPoster2?.bundle?.contentVersion) {
+      symptomPoster = symptomPoster2;
```
brunobowden, line 137: This loads the content serially rather than in parallel?

It bothers me that there's so much duplicated code from Line 91 to 137. I'm not sure if the logic can be abstracted out easily. Not a blocker on commit but please consider refactoring if it's relatively straight forward.

client/flutter/lib/pages/main_pages/check_up_poster_page.dart
```diff
@@ -0,0 +1,190 @@
...
+              data: content.bodyHtml,
+              defaultTextStyle: defaultTextStyle,
+              customTextStyle: (dom.Node node, TextStyle baseStyle) {
+                if (node is dom.Element) {
+                  switch (node.localName) {
+                    case 'b':
+                      return baseStyle.merge(boldTextStyle);
+                  }
+                }
+                return baseStyle.merge(defaultTextStyle);
```
brunobowden, line 183: The nested `return` statements took me a while to parse. I think it would read better to move parts of this before the initial return statement.... particularly L184..173... then also line 168 onwards.

client/flutter/lib/api/content/content_store.dart
advayDev1: Unfortunately its not straightforward.  Dart doesn't allow you to pass setters as functions, so its not really possible.

client/flutter/lib/pages/main_pages/check_up_poster_page.dart
advayDev1: Done

PR #1481, opened 2020-09-03
🗑️ Delete EN plugin, EN variant, Present Compiler, and S2 which are not used

advayDev1:
Keeping the License reference to Present pending confirmation from @crazybob .


brunobowden (approved):
Add in the description a note that you're intentionally keeping the Present reference in the license and will remove that later.

advayDev1:
Good call - done.

PR #1489, opened 2020-09-10
Triaged work should not be marked stale

advayDev1:


brunobowden (approved):
A very sensible change

PR #1491, opened 2020-09-10
🏡 New home promo ; 🆕 use newer content ; and more fixes

### LINKED ISSUE #1429: New home screen banner/hero mocks (without Check-Up)
Please provide a new mock for what the home page banner/hero looks like if not pointing to Check-Up tool.

### LINKED ISSUE #1452: If there's a local content bundle and a network bundle, use the one with highest content_version

### LINKED ISSUE #1490: Fetch content in parallel
Currently ContentStore updates all content in sequence.  Don't do that

advayDev1:
- Home promo per David's mocks
- Refactor the content updating to be parallel async
- Don't wait for timeout when there is no network connectivity
- Refresh content on connectivity change
- When there is a network version and a binary asset version, choose the newer content
- Let the underlying cache store data for many years, so we can still return old cached content if it is newer than the binary asset content
- Actually return new network versions
- Adds `connectivity` dep (which is run by flutter.dev, which depends on Reachability pod)

Fixes #1490 ; Fixes #1452 ; Fixes #1429

TESTED = Locally on my Android phone with various combos of network connectivity and varying versions of the home content I was modifying.

Screenshot_20200910-124159

### DIFF
--- a/client/flutter/lib/api/content/content_store.dart
+++ b/client/flutter/lib/api/content/content_store.dart
@@ -76,6 +76,15 @@ abstract class _ContentStore with Store implements Updateable {
   @observable
   PosterContent symptomPoster;
 
+  Future<void> updateBundle<T extends ContentBase>(String name, T old,
+      T Function(ContentBundle) constructor, void Function(T) setter) async {
+    final newValue =
+        constructor(await service.load(locale, countryIsoCode, name));
+    if (newValue?.bundle?.contentVersion != old?.bundle?.contentVersion) {
+      setter(newValue);
+    }
+  }
+
   @action
   Future<void> update() async {
     // TODO: UserPreferences should be injected dependency.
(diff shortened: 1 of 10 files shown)


theswerd (approved):
LGTM

client/flutter/lib/api/content/content_store.dart
```diff
@@ -76,6 +76,15 @@ abstract class _ContentStore with Store implements Updateable {
   @observable
   PosterContent symptomPoster;
 
+  Future<void> updateBundle<T extends ContentBase>(String name, T old,
+      T Function(ContentBundle) constructor, void Function(T) setter) async {
+    final newValue =
+        constructor(await service.load(locale, countryIsoCode, name));
+    if (newValue?.bundle?.contentVersion != old?.bundle?.contentVersion) {
```
hspinks, line 83: question: do we want only newer content versions, or any different content version?
```suggestion
    if (newValue?.bundle?.contentVersion > old?.bundle?.contentVersion) {
```

hspinks (approved):
lgtm + 1 non-blocking question

client/flutter/lib/api/content/content_store.dart
advayDev1: Sure we can do that, I was thinking that it is enforced in the cache manager, but this is fine too.

PR #1494, opened 2020-09-11
Prevent pop-in of home screen logo from shifting the page; and delete unused graphics

advayDev1:
We hardcode the width, why not the height?


britannio (changes requested):
The path for some of the deleted graphics is still present in `l10n.dart` however, they seem to be unused. Should they be removed?

advayDev1:
> The path for some of the deleted graphics is still present in `l10n.dart` however, they seem to be unused. Should they be removed?

Hmm, yes they should be but I'm not sure how to auto-regen those files...

client/flutter/pubspec.yaml
```diff
@@ -75,9 +75,7 @@ flutter:
     - assets/images/
     - assets/svg/
     - assets/animations/
-    - assets/news_press/
     - assets/onboarding/
-    - assets/travel_advice/
```
brunobowden, line 80: These files are referenced in the arb file file, where they aren't removed. I'm expecting for them to be removed from the arb file(s) or to explain why it should be done any other way.

https://github.com/WorldHealthOrganization/app/blob/c02ffef9a0151dbe8248dcc745efa35b859b16b9/client/flutter/lib/l10n/intl_en_US.arb#L68

brunobowden (changes requested):

brunobowden:
@advayDev1 @britannio - I think they need to be removed from the `l10n/intl_en_US.arb` file... then I'm guessing they'll be removed from all the other generated dart files.

https://github.com/WorldHealthOrganization/app/blob/c02ffef9a0151dbe8248dcc745efa35b859b16b9/client/flutter/lib/l10n/intl_en_US.arb#L68

client/flutter/pubspec.yaml
advayDev1: fixed

advayDev1:
Good call - removed all png references from loc files. PTAL @brunobowden @britannio

advayDev1:
Please merge whenever one of you approves - Thanks.

theswerd (approved):

PR #1505, opened 2020-09-21
Various fixes!

advayDev1:
1. When the server fails, don't crash the UI for stats.
2. Make all home screen cards tappable.
3.


theswerd (approved):
LGTM

PR #1514, opened 2020-09-23
Update build instructions

advayDev1:
We use cocoapod 1.9.3 and xcode likely needs an update too.


brunobowden (approved):
Nicer to do this as a precommit check but we'll leave that for a new person

advayDev1:
> Nicer to do this as a precommit check but we'll leave that for a new person

They do check this on the CI, and will break PRs if they are not followed.  Unfortunately the WHO RelEng process does not commit their intermediate files back into the repo.

PR #1549, opened 2020-10-04
Preserve context within Git about the project work done outside Git

advayDev1:


advayDev1:
@brunobowden this commit is signed with my own PGP metadata - when you commit it can you use the full merge commit (not squash or rebase).  we can test if the metadata gets preserved.

brunobowden (approved):
@advayDev1 - this will be a good test

PR #1594, opened 2020-10-10
FlatButtons => TextButtons

theswerd:
### Flutter 1.22.1 Styling Changes

- Switched all **[FlatButtons](https://api.flutter.dev/flutter/material/FlatButton-class.html)** to **[TextButtons](https://api.flutter.dev/flutter/material/TextButton-class.html)** as they are now depreciated
- Updated the Stack **overflow** to **clipBehavior** as overflow is now depreciated.
 

* [x] iOS Simulator
* [ ] Android Emulator
* [ ] iOS Device
* [ ] Android Device

---

- [ ] REQUIRED: Do you have an Issue **assigned to you** for this PR?
- [x] Verified your name is in the [docs/credits.yaml](https://github.com/WorldHealthOrganization/app/blob/master/docs/credits.yaml) file (if you want it to be).

### DIFF
--- a/client/flutter/lib/components/page_button.dart
+++ b/client/flutter/lib/components/page_button.dart
@@ -61,14 +61,34 @@ class _PageButtonState extends State<PageButton> {
 
   @override
   Widget build(BuildContext context) {
-    return FlatButton(
+    return TextButton(
       onPressed: widget.onPressed != null ? _onPressed : null,
-      disabledColor: Constants.neutralTextLightColor.withOpacity(0.4),
-      disabledTextColor: Colors.white,
-      shape: RoundedRectangleBorder(
-        borderRadius: BorderRadius.circular(widget.borderRadius),
-      ),
-      color: widget.backgroundColor,
+      style: ButtonStyle(
+          foregroundColor: MaterialStateProperty.resolveWith(
+            (states) {
+              if (states.contains(MaterialState.disabled)) {
+                return Colors.white;
+              } else {
+                return null;
+              }
+            },
+          ),
+          backgroundColor: MaterialStateProperty.resolveWith(
+            (states) {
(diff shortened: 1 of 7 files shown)


client/flutter/lib/components/page_button.dart
```diff
@@ -61,14 +61,34 @@ class _PageButtonState extends State<PageButton> {
...
+                return null;
+              }
+            },
+          ),
+          backgroundColor: MaterialStateProperty.resolveWith(
+            (states) {
+              if (states.contains(MaterialState.disabled)) {
+                return Constants.neutralTextLightColor.withOpacity(0.4);
+              } else {
+                return widget.backgroundColor;
```
brunobowden, line 81: Just a guess whether `widget.backgroundColor` is used as the default or does this need to be explicitly set?

brunobowden (approved):
@britannio - if you can look at this with more Flutter experience, that would be appreciated. Otherwise this looks good to me.

@theswerd - please add more details about whether this should be a visible or invisible change. Think of how a reviewer like me needs to evaluate it? If it's invisible, then the commit message should affirmatively state that and a screenshot for reassurance. If it's visible, then before and after screenshots. Also, this needs to be tested on Android as well given that we need to keep things safer at this point.

client/flutter/lib/components/page_button.dart
theswerd: `widget.backgroundColor` is not the default. It is passed into the `pageButton` class and then put into the button.

yukuairoy (approved):

brunobowden:
@theswerd - please rebase and let us know when it's ready for review again. Happy to take this PR but we should focus on launching first before doing any more major refactors.

PR #1619, opened 2020-10-15
App Engine and Firebase Terraform

### LINKED ISSUE #1555: App Engine and Firebase Terraform
This is to the first set of changes to cover creation of the App Engine and Firebase projects using Terraform.

Philosophically, we want to use Infrastructure as code. Beyond normal IaC benefits, it specifically helps this project in duplicating the same set up between the dev and prod projects. Provides greater transparency and reproducibility.

## Checklist:
- [x] Issue has descriptive title focused on the feature or problem from the user's perspective

yukuairoy:
Closes #1555 

How did you test the change?

- [ ] iOS Simulator
- [ ] iOS Device
- [ ] Android Simulator
- [ ] Android Device
- [ ] other, please describe

## Checklist:

### DIFF
--- a/.github/workflows/policy-check.yaml
+++ b/.github/workflows/policy-check.yaml
@@ -39,3 +39,5 @@ jobs:
       # Match the tools/.lintstagedrc.json file
       - run: npx --no-install prettier -c '../**/*.{md,java,js,sh,yaml,yml}|*.{md,java,js,sh,yaml,yml}'
         working-directory: ./tools
+      - run: terraform fmt -check -recursive .
+        working-directory: ./server/terraform
--- a/server/terraform/defaults.auto.tfvars
+++ b/server/terraform/defaults.auto.tfvars
@@ -1,0 +1,1 @@
+location_id = "europe-west6" # Switzerland is the preferred location for WHO
--- a/server/terraform/developer-guide.md
+++ b/server/terraform/developer-guide.md
@@ -1,0 +1,23 @@
+## How to create and setup GCP projects using terraform
+
+### Prerequisites
+
+- [Install Terraform](https://learn.hashicorp.com/tutorials/terraform/install-cli?in=terraform/gcp-get-started)
+- Make sure you have access to `who-terraform-admin` project.
+- [Create a new key](https://www.terraform.io/docs/providers/google/guides/getting_started.html#adding-credentials) for the `terraform@who-terraform-admin.iam.gserviceaccount.com` service account in the `who-terraform-admin` project, download the JSON credentials, and configure your environment:
+
+  ```sh
+  export GOOGLE_APPLICATION_CREDENTIALS=<path to json credentials>
+  ```
+
+### Create or update a GCP project
+
+- Change directory into `terraform/` and deploy
+
+  ```sh
+  cd server/terraform
+  terraform init
+  terraform apply -var-file=staging.tfvars
+  # to delete resource
+  terraform destroy
+  ```
--- a/server/terraform/project.tf
+++ b/server/terraform/project.tf
@@ -1,0 +1,41 @@
+variable "project_id" {
+  type = string
+}
+variable "billing_account" {
+  type = string
+}
+variable "location_id" {
+  type = string
+}
+variable "org_id" {
+  type = string
+}
+
+provider "google" {
+  region = var.location_id
+}
+
+resource "google_project" "project" {
+  provider        = google-beta
+  name            = var.project_id
+  project_id      = var.project_id
+  billing_account = var.billing_account
+  org_id          = var.org_id
+}
+
+resource "google_firebase_project" "default" {
+  provider = google-beta
+  project  = google_project.project.project_id
+}
+
+resource "google_firebase_project_location" "basic" {
+  provider    = google-beta
+  project     = google_firebase_project.default.project
+  location_id = var.location_id
+}
+
+resource "google_app_engine_application" "app" {
+  project       = google_firebase_project.default.project
+  location_id   = var.location_id
+  database_type = "CLOUD_DATASTORE_COMPATIBILITY"
+}
--- a/tools/.lintstagedrc.json
+++ b/tools/.lintstagedrc.json
@@ -1,10 +1,12 @@
 {
-  
   "../**/*.{md,java,js,sh,yaml,yml}|*.{md,java,js,sh,yaml,yml}": [
     "prettier --write"
   ],
   "../**/*.dart|*.dart": [
     "flutter format"
+  ],
+  "../server/terraform/": [
+    "terraform fmt -recursive"
   ],
   "../client/**": [
     "./verify-client.sh"
--- a/.gitignore
+++ b/.gitignore
@@ -107,3 +107,8 @@ buck-out/
 .classpath
 .project
 .settings
+
+# Terraform
+*.tfstate
+*.tfstate.backup
+server/terraform/.terraform
--- a/server/README.md
+++ b/server/README.md
@@ -121,6 +121,10 @@ And, if you want to be able to manipulate Firebase:
 
     $  gcloud components install beta app-engine-java && gcloud components update
 
+### Install Terraform
+
+Follow the directions [here](https://learn.hashicorp.com/tutorials/terraform/install-cli?in=terraform/gcp-get-started)
+
 ### Install IntelliJ IDE (Optional)
 
     $ brew cask install intellij-idea-ce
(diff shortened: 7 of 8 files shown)


terraform/developer-guide.md
```diff
@@ -0,0 +1,39 @@
...
+  ```sh
+  export TF_VAR_billing_account=<GCP billing account id>
+  ```
+
+### Create or update a GCP project
+
+- Configure environment:
+
+  ```sh
+  export TF_VAR_project_name=<project name>
```
brunobowden, line 24: I think the instances should be specified in tf-var files. This puts it under source control: https://www.terraform.io/docs/configuration/variables.html#variable-definitions-tfvars-files

So the documented approach ends up with distinct files for prod, staging, hack, dev and so on. Staging and hack should be automatic pushes. Prod and dev can be manually done. Not sure if you removed `=` then the autocomplete would work which would be nice.

I like this better as the action is more obvious and not hidden by environment variables:
```
terraform apply -var-file=staging.tfvars
```

terraform/project.tf
```diff
@@ -0,0 +1,31 @@
+variable "project_name" {}
```
brunobowden, line 1: Repeat for all variables:
```suggestion
variable "project_name" {
  type = string
}
```

terraform/project.tf
```diff
@@ -0,0 +1,31 @@
...
+variable "billing_account" {}
+variable "region" {}
+
+provider "google" {
+  region = var.region
+}
+
+resource "random_id" "id" {
+  byte_length = 4
+  prefix      = var.project_name
```
brunobowden, line 11: Randomization is to make the name unique? I'd rather not have random fluff in the names as it's then much harder to remember and type. I think using the "who-myhealth2" as a prefix should suffice. You could declare standard shared config like this in a `terraform.tfvars` file which is automatically loaded anytime Terraform is run.

brunobowden (changes requested):
@yukuairoy - this is great progress. I think we should move it to the `server/terraform` directory to keep the tree cleaner. 

Please also run `terraform fmt` just to check formatting. You can automatically enforce that later in #1620

brunobowden:
@yukuairoy - you don't need to do everything here. Main thing, is to add something to description about testing:

Testing:
1. Deployed App Engine using modified script (I wonder if this should be defined within terraform or not?)
1. Tested App Engine instead with curl tests

terraform/project.tf
```diff
@@ -0,0 +1,31 @@
...
+}
+
+resource "google_project" "project" {
+  provider        = google-beta
+  name            = var.project_name
+  project_id      = random_id.id.hex
+  billing_account = var.billing_account
+  org_id          = 532343229286
```
epicfaace, line 19: Would be best to make this a variable as well

terraform/project.tf
```diff
@@ -0,0 +1,31 @@
+variable "project_name" {}
+variable "billing_account" {}
```
epicfaace, line 2: It's also best practice to put all variables into a separate `vars.tf` file -- see https://github.com/cisagov/crossfeed/blob/master/infrastructure/vars.tf for an example.

terraform/developer-guide.md
```diff
@@ -0,0 +1,39 @@
...
+  export TF_VAR_region=<valid GCP region, e.g. europe-west6>
+  ```
+
+- Change directory into `terraform/` and do
+  ```sh
+  terraform init
+  ```
+- Deploy
+  ```sh
+  terraform apply
```
epicfaace, line 34: It's probably better practice to do the following (adding `out` to `.gitignore`, of course):

```
terraform plan -out out
terraform apply out
```

so the user is exactly sure about which changes will happen before they actually get applied

brunobowden: @epicfaace - Terraform by default shows you the changes plan before it's applied. I've used this many times and it's worked fine for me. The only thing that I think would be dangerous is `-auto-approve`:

```
terraform apply -auto-approve          Skip interactive approval of plan before applying.
```

brunobowden:
@yukuairoy - I think it needs to specify `CLOUD_DATASTORE_COMPATIBILITY` to create the old Datastore database type upon initial creation: https://www.terraform.io/docs/providers/google/r/app_engine_application.html#database_type

server/terraform/project.tf
```diff
@@ -0,0 +1,41 @@
...
+resource "google_firebase_project_location" "basic" {
+  provider    = google-beta
+  project     = google_firebase_project.default.project
+  location_id = var.region
+}
+
+resource "google_app_engine_application" "app" {
+  project       = google_firebase_project.default.project
+  location_id   = var.region
+  database_type = "CLOUD_FIRESTORE"
```
brunobowden, line 40: Needs to be: `CLOUD_DATASTORE_COMPATIBILITY` until it's upgraded

server/terraform/developer-guide.md
```diff
@@ -0,0 +1,29 @@
+## How to create and setup GCP projects using terraform
+
+### Prerequisites
+
+- [Install Terraform](https://learn.hashicorp.com/tutorials/terraform/install-cli?in=terraform/gcp-get-started)
```
brunobowden, line 5: Update `server/README.md` to add the terraform install step. Devs will need this for server work to validate changes even if they don't edit the Terraform files.

server/terraform/project.tf
```diff
@@ -0,0 +1,41 @@
+variable "project_name" {
+  type = string
+}
+variable "billing_account" {
+  type = string
+}
+variable "region" {
```
brunobowden, line 7: Should this be `location_id` to match the convention for Google Cloud? Or is this a convention for Terraform?

server/terraform/project.tf
```diff
@@ -0,0 +1,41 @@
+variable "project_name" {
```
brunobowden, line 1: Should this be `project_id` to match the convention for Google Cloud? Or is this a convention for Terraform?

.github/workflows/policy-check.yaml
```diff
@@ -39,3 +39,5 @@ jobs:
       # Match the tools/.lintstagedrc.json file
       - run: npx --no-install prettier -c '../**/*.{md,java,js,sh,yaml,yml}|*.{md,java,js,sh,yaml,yml}'
         working-directory: ./tools
+      - run: terraform fmt -check '../**/*.tf|*.tf'
```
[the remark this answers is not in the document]
brunobowden: Clearly this failed the CI. Try running this locally to see what works:
https://github.com/WorldHealthOrganization/app/pull/1619/checks?check_run_id=1283328236

tools/.lintstagedrc.json
```diff
@@ -6,8 +6,8 @@
   "../**/*.dart|*.dart": [
     "flutter format"
   ],
-  "../**/*.tf|*.tf": [
-    "terraform fmt"
+  "../": [
```
brunobowden, line 9: I just ran it in the top level folder and it seemed to take a second to run. A little annoying to add that cost to everyone. I think this only runs it when there's a change that overlaps with the terraform directory:
```suggestion
  "../server/terraform/**": [
```

brunobowden (approved):
Just one minor issue and I think we're good to go. Feel free to defer the `terraform validate` to another PR.

server/terraform/defaults.auto.tfvars
```diff
@@ -0,0 +1 @@
+location_id = "europe-west6"
```
brunobowden, line 1: Add comment:
```
# Switzerland is the preferred location for WHO
```

.github/workflows/policy-check.yaml
```diff
@@ -39,3 +39,5 @@ jobs:
       # Match the tools/.lintstagedrc.json file
       - run: npx --no-install prettier -c '../**/*.{md,java,js,sh,yaml,yml}|*.{md,java,js,sh,yaml,yml}'
         working-directory: ./tools
+      - run: terraform fmt -check -recursive './terraform/'
+        working-directory: ./server
```
brunobowden, line 43: Don't split the working directory split across two lines:
```suggestion
      - run: terraform fmt -check -recursive .
        working-directory: ./server/terraform
```

tools/.lintstagedrc.json
```diff
@@ -1,15 +1,9 @@
...
-  "../**/*.dart|*.dart": [
-    "flutter format"
-  ],
-  "../client/**": [
-    "./verify-client.sh"
-  ],
-  "../server/**": [
-    "./verify-server.sh"
-  ]
+  "../**/*.dart|*.dart": ["flutter format"],
```
brunobowden, line 5: Did you run prettier on this file? I'm fine with removing the empty line 2 but I think the rest of it works better with the original layout.

brunobowden (commented):
@yukuairoy - please add some comments about testing as well.

brunobowden:
@yukuairoy just told me over a phone call that this was good to merge in

PR #1652, opened 2020-10-22
Propagate relicense of madvay/actions-sandbox

advayDev1:
For ease of the upcoming attempted relicense I've relicensed my underlying repository linked here to Apache 2.

The only people who seem to have worked on the files I pulled over into this repo seem to be @brunobowden and @hspinks .  If you both agree to this relicense by saying "I agree" I will remove the VOID and Bruno can submit.
As always this is not legal advice and I cannot represent that this is legally sufficient to perform the relicense.

Please note that if you don't submit this now, it is possible others (other than Hunter and Bruno) will further adapt my sandbox-licensed work and you may need more agreements here.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -119,24 +119,16 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 --------------------------------------------------------------------------------
 For content from: https://github.com/madvay/actions-sandbox/
 
-MIT License
-
 Copyright (c) 2020 Advay Mengle
 
-Permission is hereby granted, free of charge, to any person obtaining a copy
-of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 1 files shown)


brunobowden:
I agree. Thank you Advay

brunobowden:
@advayDev1 - was this just used as an experimental sandbox? My impression was that there wasn't any dependencies on `actions-sandbox`. If there were, then I'd be inclined to bring the code in to WHO/app repo with your permission but I don't think we need to.

brunobowden (approved):

advayDev1:
I directly copied workflows and scripts into this codebase, hence the copyright notice.  There is no dependency intentionally.

hspinks:
I agree

advayDev1:
thanks hunter over to @brunobowden

brunobowden:
Thanks @advayDev1

Issue #1511, opened 2020-09-22
Design Review issues


advayDev1:
from dave/hunter/me review Sept 21:

- TODO(bruno): Location sharing copy needs to be replaced
- Move country selector up near text, Keep buttons aligned across screens, Align onboarding screens as: 60% for top of onboarding, 40% for the button/list section
- Tighten letter spacing for onboarding header
- Sentence case everything except COVID-19 Response
- TODO(Ayman): content - One sentence for wearing masks
- Home screen cards should be tappable to modal page
- Donate SVG Feet alignment to bottom of home/top of tabs
- Check-up poster TODO (awaiting content from Bruno, Hunter has finished the styling and structure)
- Drop shadow 1-2px blur “Busting myths...”
- P2 - show iphone scale screen on ipad
- Make GTF fully scrollable without learn more button
- TODO(Ayman): content - Travel advice updates
- TODO(Ayman): content - Protect Yourself mask instructions (full view)
- Add “Debug Info” heading to expanded About the App View
- TODO(future): make the stats graphs scrolling 30-day windows

brunobowden:
FYI @theswerd - see if there's any essential issues to be addressed out of this.

brunobowden:
@advayDev1 - please could you edit my description. It appeared that a lot of these are done.... I just could discern the status as TODO != DONE.

advayDev1:
I think you categorized them correctly but note that we put all medical content as todo-ayman since WHO has to approve all of it.

brunobowden:
@advayDev1 - the whole app needs to go through medical approval. There's also discussion of changes over email too. What else was highest priority? If it's critical for launch, then I'd like to separate it out in to individual issues. The GTF Scrollability is separate out but not anything else at this point.

Documenting original description for reference:

From dave/hunter/advay review Sept 21:
1. TODO(bruno): Location sharing copy needs to be replaced
2. Move country selector up near text, Keep buttons aligned across screens, Align onboarding screens as: 60% for top of onboarding, 40% for the button/list section
3. Tighten letter spacing for onboarding header
4. Sentence case everything except COVID-19 Response
5. TODO(Ayman): content - One sentence for wearing masks
6. Home screen cards should be tappable to modal page - DONE
7. Donate SVG Feet alignment to bottom of home/top of tabs
8. Check-up poster TODO (awaiting content from Bruno, Hunter has finished the styling and structure) - DONE
[...]

PR #1660, opened 2020-10-24
Learn More Headline Opaque Shadow

### LINKED ISSUE #1571: "Get the Facts" accessibility in Learn More page
In the learn section, get the fact text and button are not very visible over the background image

theswerd:
Added blurred text behind the subtitle text

Related to suggestion in #1511

Closes #1571 

## Screenshots

<details>

Screenshot

</details>

- [x] iOS Simulator
- [ ] iOS Device
- [x] Android Simulator
- [ ] Android Device
- [ ] other, please describe

## Checklist:

### DIFF
--- a/client/lib/components/learn_page_promo.dart
+++ b/client/lib/components/learn_page_promo.dart
@@ -1,3 +1,5 @@
+import 'dart:ui';
+
 import 'package:flutter/cupertino.dart';
 import 'package:flutter_svg/flutter_svg.dart';
 import 'package:who_app/api/linking.dart';
@@ -5,6 +7,18 @@ import 'package:who_app/components/button.dart';
 import 'package:who_app/components/promo_curved_background.dart';
 import 'package:who_app/components/themed_text.dart';
 import 'package:who_app/constants.dart';
+
+const shadow = Shadow(
+  blurRadius: 20.0,
+);
+
+// 4X shadows is used to make it more opaque and offset from the background
+const opaqueShadow = <Shadow>[
+  shadow,
+  shadow,
+  shadow,
+  shadow,
+];
 
 class LearnPagePromo extends StatelessWidget {
   final String title;
@@ -51,6 +65,7 @@ class LearnPagePromo extends StatelessWidget {
                   textAlign: TextAlign.center,
                   style: TextStyle(
                     color: CupertinoColors.white,
+                    shadows: opaqueShadow,
                   ),
                 ),
                 ThemedText(
@@ -59,6 +74,7 @@ class LearnPagePromo extends StatelessWidget {
                   textAlign: TextAlign.center,
                   style: TextStyle(
                     color: CupertinoColors.white,
+                    shadows: opaqueShadow,
                   ),
                 ),
                 SizedBox(


advayDev1:
This won't solve the accessibility problem.  The idea was a blur of the text itself.  Like white text, on blurred black version of the same text immediately behind it.  That creates a higher contrast ratio for clarity.

yukuairoy (approved):

theswerd:
@advayDev1 Done 

simulator_screenshot_51238CE1-4B13-4190-9849-9B06A430FE0C

brunobowden:
@advayDev1 is right. See example no. 9 ont his page which is I believe what he’s suggesting and what I’d consider with: https://www.slideteam.net/blog/11-hacks-to-make-text-over-images-more-readable-craft-a-stunning-slide

The current blurring looks a little strange like it might be a mistake. If the blurred black text is too difficult then we’ll reconsider.

advayDev1:
@brunobowden @theswerd try this. https://stackoverflow.com/a/52860786 its built into flutter these days

theswerd:
@advayDev1 alr done

brunobowden:
@theswerd - please update the PR and give us a chance to review

theswerd:
@brunobowden new attempt:
screenshot demo

brunobowden:
@theswerd - if the no offset version looks very similar, then it would be preferable to use that as the code is so much simpler. There's significant value in simpler code. Also, please update the screenshot in the description. Then when you next push / comment mention "Description updated with new screenshot".

theswerd:
@brunobowden description updated

client/lib/components/learn_page_promo.dart
```diff
@@ -6,6 +8,22 @@ import 'package:who_app/components/promo_curved_background.dart';
 import 'package:who_app/components/themed_text.dart';
 import 'package:who_app/constants.dart';
 
+const double blurRadius = 20;
+const List<Shadow> headerTextShadows = [
```
brunobowden, line 12: Can we make it less verbose? Something like:
```suggestion
const shadow = Shadow(blurRadius: blurRadius);
// 4X shadows is used to make it more opaque and offset from the background
const opaqueShadow = <Shadow>[
   shadow,
   shadow,
   shadow,
   shadow]
```
The reason to use `opaqueShadow` is that it more closely matches the function. You could imagine using this on other text.... with a constant that's shared more widely on other pages.

See Dart type system for preference in specifying type on the `[]`:
https://dart.dev/guides/language/type-system

client/lib/components/learn_page_promo.dart
```diff
@@ -6,6 +8,22 @@ import 'package:who_app/components/promo_curved_background.dart';
 import 'package:who_app/components/themed_text.dart';
 import 'package:who_app/constants.dart';
 
+const double blurRadius = 20;
```
brunobowden, line 11: Type is automatically inferred by appending `.0` on number:
```suggestion
const blurRadius = 20.0;
```

brunobowden (approved):

client/lib/components/learn_page_promo.dart
```diff
@@ -6,6 +8,17 @@ import 'package:who_app/components/promo_curved_background.dart';
 import 'package:who_app/components/themed_text.dart';
 import 'package:who_app/constants.dart';
 
+const blurRadius = 20.0;
+const shadow = Shadow(blurRadius: blurRadius);
```
brunobowden, line 12: I'd remove the `const blurRadius = 20.0;` and just inline it since the value isn't used anywhere else.
```suggestion
const shadow = Shadow(blurRadius: 20.0);
```

brunobowden:
@theswerd - one minor point of feedback, please correct that if you agree, squash and merge at your leisure

client/lib/components/learn_page_promo.dart
```diff
@@ -9,7 +9,7 @@ import 'package:who_app/components/themed_text.dart';
 import 'package:who_app/constants.dart';
 
 const blurRadius = 20.0;
```
brunobowden, line 11: Remove prior line

PR #1661, opened 2020-10-24
Bar Graph Colors

theswerd:
Add indicator on bar chart

## Screenshots

<details>

Screenshot 1
Screenshot 2
Screenshot 3
Screenshot 4
Demo

</details>

- [x] iOS Simulator
- [ ] iOS Device
- [x] Android Simulator
- [ ] Android Device
- [ ] other, please describe

## Checklist:

### DIFF
--- a/client/lib/components/recent_numbers_graph.dart
+++ b/client/lib/components/recent_numbers_graph.dart
@@ -1,5 +1,6 @@
 import 'package:fl_chart/fl_chart.dart';
 import 'package:flutter/cupertino.dart';
+import 'package:flutter/material.dart';
 import 'package:intl/intl.dart';
 import 'package:who_app/components/themed_text.dart';
 import 'package:who_app/constants.dart';
@@ -7,15 +8,20 @@ import 'package:who_app/pages/main_pages/recent_numbers.dart';
 import 'package:who_app/proto/api/who/who.pb.dart';
 
 const double padding = 30;
+const Duration defaultSwapDuration = Duration(
+  milliseconds: 750,
+);
 
 class RecentNumbersBarGraph extends StatefulWidget {
   final DataAggregation aggregation;
   final DataDimension dimension;
+  final Color graphColor;
   const RecentNumbersBarGraph({
     Key key,
-    this.timeseries,
-    this.aggregation,
-    this.dimension,
+    @required this.timeseries,
+    @required this.aggregation,
+    @required this.dimension,
+    @required this.graphColor,
   }) : super(key: key);
 
   final List<StatSnapshot> timeseries;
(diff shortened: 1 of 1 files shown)


client/lib/components/recent_numbers_graph.dart
```diff
@@ -40,6 +54,19 @@ class _RecentNumbersBarGraphState extends State<RecentNumbersBarGraph> {
           show: false,
         ),
         barTouchData: BarTouchData(
+          touchCallback: (touchResponse) {
+            setState(() {
+              swapDuration = Duration(microseconds: 1);
```
brunobowden, line 59: Why 1000 Millisecs here and 750 above? Ideally the duration should be specified once only.

More importantly, why change the value at all? I’d expect there to be a single default and not to need changing it anywhere as you also do below.

https://en.m.wikipedia.org/wiki/Don%27t_repeat_yourself

client/lib/components/recent_numbers_graph.dart
```diff
@@ -40,6 +54,19 @@ class _RecentNumbersBarGraphState extends State<RecentNumbersBarGraph> {
...
+              if (touchResponse.spot != null &&
+                  touchResponse.touchInput is! FlPanEnd &&
+                  touchResponse.touchInput is! FlLongPressEnd) {
+                selectedIndex = touchResponse.spot.touchedRodDataIndex;
+              } else {
+                selectedIndex = null;
+                swapDuration = Duration(milliseconds: 750);
```
brunobowden, line 66: Should be using default and not need adjustment

client/lib/components/recent_numbers_graph.dart
```diff
@@ -109,7 +140,9 @@ class _RecentNumbersBarGraphState extends State<RecentNumbersBarGraph> {
         bars.add(
           BarChartRodData(
             y: 0,
-            color: Constants.textColor.withOpacity(.3),
+            color: Constants.textColor.withOpacity(
+              .3,
+            ),
```
brunobowden, line 145: Style change above is allowed due to editing code. This section shouldn’t be changed as you’re not changing anything.

brunobowden (approved):
Thanks Ben. The animated screenshot is very helpful.

This is optional but should the colored bar be a little wider? How visible is it right now on small mobile devices? I don’t have a strong opinion on this but something we should consider in QA. FYI @Mohit020888

client/lib/components/recent_numbers_graph.dart
theswerd: The duration for animating the swap between **Daily** and **Total Stats** is 750 milliseconds, however the animation for changing the bars should be instantaneous. This is here so after the bar color changes are done, the animation returns to its original state.

client/lib/components/recent_numbers_graph.dart
theswerd: Originally had a change on that line -- will revert now

client/lib/components/recent_numbers_graph.dart
brunobowden: Sorry, misread that... it's 0.001 milliseconds for a microsecond. That's fine then.

brunobowden (commented):
Extra 750 milliseconds to shared constant?

theswerd:
@brunobowden done -- 750 milliseconds has been refactored

brunobowden:
@theswerd - looks good, merge at your leisure

PR #1665, opened 2020-10-25
TOS and PN links moved to WHO website

### LINKED ISSUE #1510: Privacy Notice / TOS published and linked
Per convo with WHO/Joris today: 
- WHO to verify the Privacy Policy/ToS are in all the right places in app stores, etc. prior to download (if WHO deems that necessary)
- WHO to verify its comfort with any data that might leave the device/app prior to passing past the first ToS/Privacy screen (see, for example, but not only, https://github.com/WorldHealthOrganization/app/issues/1509 , https://github.com/WorldHealthOrganization/app/issues/934 , https://github.com/WorldHealthOrganization/app/issues/1308 )

Obviously this is not legal advice to WHO, and WHO needs to figure out how it wants to do things.

cc @brunobowden 

_Please don't edit to signify changes - GitHub makes it appear that such edits come from me.  If your goal is different than what is described here, say so in comments._

brunobowden:
Close #1510

Links updated:
- https://whoapp.org/terms => https://www.who.int/myhealthapp/terms-of-use
- https://whoapp.org/privacy => https://www.who.int/myhealthapp/privacy-notice
- Remove TOS and Privacy policies from server

Correction:
- aboutPagetermsOfServiceLinkUrl => aboutPagePrivacyPolicyLinkUrl

- [ ] iOS Simulator
- [ ] iOS Device
- [x] Android Simulator - TOS & PN links on Install.... and Settings page
- [ ] Android Device
- [ ] other, please describe

## Checklist:


server/appengine/src/main/java/who/WhoServletModule.java
```diff
@@ -33,8 +33,6 @@ protected void configureServlets() {
     serve("/remote_api").with(new RemoteApiServlet());
 
     serve("/app").with(new AppStoreServlet());
-    serve("/terms").with(forwardTo("/terms.pdf"));
```
advayDev1, line 36: Old versions of the app will still point to these URLs.  Do you want to HTTP 301 them to the new urls?

brunobowden: If the app was already released then we'd have to do this. Given the small number of users, that they've already been provided with a working link at the time they accepted the TOS and PN.... and that we'll be pushing a new build first thing next week, I'd rather focus the limited resources on more pressing issues.

brunobowden: Advay - the other thought is that as soon as you push the Static Content Service, the old apps should pick up the new location and show that in the UI, so they shouldn't ever need a redirect. That'll make for a nice test in a way.

advayDev1: I'm not sure what the Static Content Service has to do with this though.  That is for the YAMLs, not for the onboarding/ToS screens which are coded.

brunobowden: Fair point but I'll hold to the original logic. What I'll probably do is push the new binary, then wait a day for people to upgrade and then switch out the server.

brunobowden:
@yukuairoy - this needs your approval as a server change

yukuairoy (approved):

PR #1711, opened 2020-11-05
Stale Content Warning

theswerd:
Added UI indicating if a cache could be outdated

## Screenshots

<details>

Simulator Screen Shot - iPhone 12 Pro Max - 2020-11-04 at 16 41 18
Simulator Screen Shot - iPhone 12 Pro Max - 2020-11-04 at 16 41 15

</details>

- [x] iOS Simulator
- [ ] iOS Device
- [ ] Android Simulator
- [ ] Android Device
- [ ] other, please describe

## Checklist:

### DIFF
--- a/client/lib/api/content/content_store.dart
+++ b/client/lib/api/content/content_store.dart
@@ -152,7 +152,7 @@ abstract class _ContentStore with Store implements Updateable {
         ),
       );
       print(
-        'ContentStore update failed',
+        'ContentStore update failed:\n$e',
       );
     }
   }
--- a/client/lib/api/content/content_store.dart
+++ b/client/lib/api/content/content_store.dart
@@ -147,7 +147,7 @@ abstract class _ContentStore with Store implements Updateable {
       );
     } catch (e) {
       unawaited(
-        FirebaseCrashlytics.instance.log(e),
+        FirebaseCrashlytics.instance.recordFlutterError(e),
       );
       print(
         'ContentStore update failed',
--- a/client/lib/api/content/content_store.dart
+++ b/client/lib/api/content/content_store.dart
@@ -11,7 +11,7 @@ import 'package:who_app/api/content/schema/question_content.dart';
 import 'package:who_app/api/display_conditions.dart';
 import 'package:who_app/api/updateable.dart';
 import 'package:who_app/api/user_preferences.dart';
-
+import 'package:pedantic/pedantic.dart';
 // Whenever this file is modified, regenerate the .g.dart file using the command:
 // flutter packages pub run build_runner build
 part 'content_store.g.dart';
@@ -134,5 +134,10 @@ abstract class _ContentStore with Store implements Updateable {
         symptomPoster = v;
       })
     ]);
+
+    unawaited(
+      UserPreferences()
+          .setLastUpdatedContent(DateTime.now().millisecondsSinceEpoch),
+    );
   }
 }
--- a/client/lib/api/content/content_store.dart
+++ b/client/lib/api/content/content_store.dart
@@ -147,12 +147,10 @@ abstract class _ContentStore with Store implements Updateable {
       );
     } catch (e) {
       unawaited(
-        FirebaseCrashlytics.instance.log(
-          e,
-        ),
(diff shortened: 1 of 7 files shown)


yukuairoy (approved):

client/lib/api/content/content_store.dart
```diff
@@ -134,5 +134,10 @@ abstract class _ContentStore with Store implements Updateable {
         symptomPoster = v;
       })
     ]);
+
+    unawaited(
```
brunobowden, line 138: I had to read up on unawaited but this looks right. I think it's worth adding a log here.

client/lib/api/content/content_store.dart
```diff
@@ -134,5 +134,10 @@ abstract class _ContentStore with Store implements Updateable {
...
+      UserPreferences()
```
brunobowden, line 139: Does this wait on the CaseStats being updated as well? At the moment, it looks as though the yaml could updated... at which point the client doesn't want about old data, even though the CaseStats could be stale? It needs to track the oldest data from either system.

brunobowden (commented):
Initial feedback but still need to go through the rest of the code

client/lib/api/content/content_store.dart
theswerd: This currently looks at only the overall last time the stats were updated.  In the next round I will extend it to each individual Store.

theswerd: Its also worth noting that when the stats store and content store update, they update by calling the overall update function.

[Stats Store](https://github.com/WorldHealthOrganization/app/blob/f3080c4c7ccf5cbaf32a12a26d76a672c7848e57/client/lib/api/stats_store.g.dart#L74)

[Content Store](https://github.com/WorldHealthOrganization/app/blob/29ec7ceba52e7f90ed9243071fbda987a892b204/client/lib/api/content/content_store.g.dart#L160)

client/pubspec.yaml
```diff
@@ -53,6 +53,7 @@ dependencies:
   # TODO: upgrade to ^0.6.5+2, build warning treated as error
   share: 0.6.3+6
   shared_preferences: ^0.5.6
+  timeago: ^2.0.28
```
brunobowden, line 56: Much better to use a library....

brunobowden (changes requested):

client/pubspec.yaml
theswerd: I've used this before, the code is relatively simple but it works

client/lib/components/page_scaffold/page_scaffold.dart
```diff
@@ -138,66 +156,88 @@ class _AlertsWrapperState extends State<AlertsWrapper>
...
+
+  void checkContentUpdate() async {
+    var lastUpdatedMilliSeconds =
+        await UserPreferences().getLastUpdatedContent();
+    if (lastUpdatedMilliSeconds == null) {
+      return;
+    }
+    lastUpdate = LastUpdate(
+        DateTime.fromMillisecondsSinceEpoch(lastUpdatedMilliSeconds));
+    if (lastUpdate.olderThan4Hours) {
```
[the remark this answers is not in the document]
theswerd: I think we would still declare stale hours within the `last_update.dart` file, as the methods related to an object specifically should be done on the object.

yukuairoy (approved):

client/lib/api/content/content_store.dart
```diff
@@ -100,39 +101,59 @@ abstract class _ContentStore with Store implements Updateable {
...
+        UserPreferences().setLastUpdatedContent(
+          DateTime.now().millisecondsSinceEpoch,
+        ),
+      );
+      print(
+        'ContentStore update finished',
+      );
+    } catch (e) {
+      unawaited(
+        FirebaseCrashlytics.instance.log(
```
brunobowden: https://firebase.google.com/docs/crashlytics/customize-crash-reports?platform=android#log-excepts
```suggestion
        FirebaseCrashlytics.getInstance().recordException(e);
```

brunobowden (approved):

Issue #1155, opened 2020-04-29
Android app does not use proxy settings for network connection


marcmcl:
Although there is a disclaimer when specifying a proxy connection that not all apps will use it, the Android App (0.6.1) will not use it.  This may cause problems for people behind both corporate firewalls and national firewalls.

advayDev1:
All HTTP requests should use the default proxy.

marcmcl:
To verify this, manually specify your IP address settings in your configuration for your network profile.  Use a proxy on your local network but specify an invalid gateway which has no Internet access.  Web browsers should work as should apps that honour proxy access.

Happy to test but be aware I'm on UK time.

marcmcl:
I've done some more investigation and found that, as mentioned in this article, both flutter and dart do not use the system-wide proxy :

https://flutterigniter.com/debugging-network-requests/

On the Android side at least it should be possible by using a java plugin as mentioned in this stackoverflow post :

https://stackoverflow.com/questions/50082372/how-do-i-get-the-proxy-settings-of-an-android-ios-device-using-flutter/50092642#50092642

Not sure how easy it would be to implement this in time for v1.0 launch.

brunobowden:
@marcmcl - would you be able to help get this in to v1? We're short staffed at the moment, so the help would be much appreciated. If not, then I think would get bumped for a future release.

marcmcl:
@brunobowden I shall give it a go.  Be aware I think Dart only supports unauthenticated proxies.

brunobowden:
@marcmcl - I think you did some work on this. Can you post the pull request so we can look it over?

marcmcl:
@brunobowden I have submitted a pull request

PR #1727, opened 2020-11-09
Marcmcl proxysupport

marcmcl:
## Screenshots

<details>
</details>

- [ ] iOS Simulator
- [ ] iOS Device
- [ ] Android Simulator
- [X] Android Device
- [ ] other, please describe
Tested by manually specifying proxy host and port, as well as with a proxy.pac URL.

## Checklist:


brunobowden:
@marcmcl - have a look at the check failures on GitHub:
1. pubspec.lock changes are missing. This is breaking most of the CI builds.
1. Podfile.lock changes are missing. Even though you're not fixing up iOS, you need to include this.

brunobowden:
@marcmcl - two further comments:
1. Your name shouldn't be in the title. It should be targeted such as "Proxy Support for Android"
1. PR description should mention associated issue: "Partial fix for #1155"
1. The [GitHub example](https://github.com/marcmcl/get_proxy#usage) is formatted incorrectly and needs this type of quote:
```
code example
```
1. What other libraries did you look at for http proxies? There seem to be a few on pub.dev. My concern is that by creating a new library, you become responsible for maintaining it long in to the future. I'd first carefully check if there's other good libraries we can use. If we have to build our own, then it might be better to do it under the WHO repo so it has more shared responsibility.

marcmcl:
@brunobowden I added the missing files and have also included my plugin files as well.  Managed to get hostname and port working for proxy on IOS too.  Still no support for proxy authentication on both IOS and Android implementations.  IOS implementation also doesn't autoconfig proxies specified via PAC yet.

brunobowden (commented):
Hi @marcmcl, thanks for contributing this. It'll take me more time to look at this. Here's some initial cursory observations.

brunobowden:
@marcmcl - there's also a conflict in `client/ios/Podfile.lock`. Not sure what to do in GitHub desktop but from the command line you'll need to:
```
git fetch upstream
git rebase -i upstream/master
```

client/lib/api/who_service.dart
```diff
@@ -93,4 +94,15 @@ class WhoService {
     }
     return 'WEB';
   }
+
+  static Future<http.Response> httpPost(dynamic url, Map<String,String> headers, dynamic body) async {
+    var httpClient = io.HttpClient();
+    var proxy = await GetProxy.getProxy;
+    if(proxy!='') {
```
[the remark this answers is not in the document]
marcmcl: @brunobowden I am still trying to work out why the automatic checks are not running on Mac.  They were running on Windows.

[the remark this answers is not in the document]
marcmcl: @brunobowden Formatting now fixed.  Check is now passing.

marcmcl:
@brunobowden going to retest tomorrow then I will mark the PR as ready for review.

client/lib/api/who_service.dart
```diff
@@ -4,7 +4,8 @@ import 'package:who_app/main.dart';
 import 'package:who_app/api/user_preferences.dart';
 import 'package:http/http.dart' as http;
 import 'dart:io' as io;
-
+import 'package:get_proxy/get_proxy.dart';
```
brunobowden, line 7: Instead of `get_proxy`, should it be `http_proxy`? You're not getting a proxy but using it for requests. It should match the `http.post` API as much as possible. The used class can then be `httpProxy` which more closely matches.

client/lib/api/who_service.dart
```diff
@@ -24,7 +25,7 @@ class WhoService {
     final postBody = jsonEncode(req.toProto3Json());
 
     final url = '$serviceUrl/putDeviceToken';
-    final response = await http.post(url, headers: headers, body: postBody);
+    final response = await httpPost(url, headers, postBody);
```
brunobowden, line 28: General thought is that the `http.post` APIs will have a lot of thought that's gone in to them. So my suggestion was to use the same API for `httpPost`. It also benefits users as they can use the same syntax and it benefits reviewers as there's fewer changes switching from one to the other. So you'd expect the new line to be:

```suggestion
    final response = await await httpProxy.post(url, headers: headers, body: postBody);
```

brunobowden (changes requested):
@marcmcl - thanks for pressing on with this. Sorry to ask you to rename lots of file as that's a painful process. Feel free to discuss / push back on the name. We should be thoughtful about that so we don't need to change it later on.

There's also two further client fetches using `webHelper.downloadFile` in `caching.dart`. Though should also use the proxy but I'm not quite sure how to do it:
https://github.com/WorldHealthOrganization/app/blob/master/client/lib/api/content/caching.dart#L44

client/lib/api/who_service.dart
marcmcl: Just a heads up that get_proxy was literally just getting the proxy information.  But I'll move the http.post alternative into this package and rename it as you suggest @brunobowden .

PR #1736, opened 2020-11-13
Properly set firebase auto init option when notifications / analytics settings are changed

### LINKED ISSUE #1308: When both analytics and notifications are turned off, set _firebaseMessaging.setAutoInitEnabled() to false
[_firebaseMessaging.setAutoInitEnabled](https://developers.google.com/android/reference/com/google/firebase/messaging/FirebaseMessaging#public-void-setautoinitenabled-boolean-enable):

> When enabled, Firebase Cloud Messaging generates a registration token on app startup if there is no valid one and generates a new token when it is deleted (which prevents deleteInstanceId() from stopping the periodic sending of data). This setting is persisted across app restarts and overrides the setting specified in your manifest.

This value is set to true after onboarding, because analytics is enabled by default. However, we need to:
- set this value to false when a user disables both analytics and notifications
- set this value to true when a user enables either analytics or notifications (or both)

epicfaace:
Properly set firebase auto init option when notifications / analytics settings are changed. Fixes #1308

`firebaseMessaging.setAutoInitEnabled(false)` is called whenever both notifications and analytics are turned off. Otherwise, when either of them are turned on, `firebaseMessaging.setAutoInitEnabled(true)` is called.

Logged to the console whenever `setAutoInitEnabled` was called and then manually tested on the iOS simulator to ensure it is called correctly in various scenarios.

- [x] iOS Simulator
- [ ] iOS Device
- [ ] Android Simulator
- [ ] Android Device
- [ ] other, please describe

## Checklist:

### DIFF
--- a/client/lib/api/user_preferences.dart
+++ b/client/lib/api/user_preferences.dart
@@ -1,6 +1,7 @@
 import 'dart:math';
 
 import 'package:firebase_analytics/firebase_analytics.dart';
+import 'package:firebase_messaging/firebase_messaging.dart';
 import 'package:shared_preferences/shared_preferences.dart';
 
 import 'package:uuid/uuid.dart';
@@ -81,8 +82,10 @@ class UserPreferences {
       await analytics.resetAnalyticsData();
     }
     await analytics.setAnalyticsCollectionEnabled(value);
-    return (await SharedPreferences.getInstance())
+    var result = (await SharedPreferences.getInstance())
         .setBool(UserPreferenceKey.AnalyticsEnabled.toString(), value);
+    await _updateFirebaseAutoInit();
+    return result;
   }
 
   Future<bool> getNotificationsEnabled() async {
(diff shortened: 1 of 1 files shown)


yukuairoy (approved):

client/lib/api/user_preferences.dart
```diff
@@ -92,8 +95,26 @@ class UserPreferences {
...
         .setBool(UserPreferenceKey.NotificationsEnabled.toString(), value);
+    await _updateFirebaseAutoInit();
+    return result;
+  }
+
+  Future<void> _updateFirebaseAutoInit() async {
+    // Updates the firebase auto init option. This method is called
+    // after a user modifies either analytics or notifications.
+    var analyticsEnabled = await getAnalyticsEnabled();
+    var notificationsEnabled = await getNotificationsEnabled();
```
brunobowden, line 108: @epicfaace - this compresses 6 lines of logic down to two. I think it's also much easier to read and debug:

```suggestion
    var firebase_auto_init = (await getNotificationsEnabled()) || (await getAnalyticsEnabled());
    // Enable Firebase Auto Init iff either notifications or analytics is enabled
    await firebaseMessaging.setAutoInitEnabled(firebase_auto_init);
```

brunobowden (approved):
@epicfaace - please make the changes before merging in. Thanks for the help.

client/lib/api/user_preferences.dart
```diff
@@ -92,8 +95,26 @@ class UserPreferences {
   }
 
   Future<bool> setNotificationsEnabled(bool value) async {
```
brunobowden, line 97: @epicfaace - do you know if this is called when someone updates the system settings? For example if they disable notifications in the iOS system settings... but leave it enabled within the app, it would be confusing. If that's the case, please file a bug on it. Not a priority to fix for v1 but I think something we should track for the future.

epicfaace: I don't think so -- will file a bug

epicfaace: See https://github.com/WorldHealthOrganization/app/issues/1741

client/lib/api/user_preferences.dart
```diff
@@ -92,8 +95,21 @@ class UserPreferences {
...
+
+  Future<void> _updateFirebaseAutoInit() async {
+    // Updates the firebase auto init option. This method is called
+    // after a user modifies either analytics or notifications.
+    var firebaseMessaging = FirebaseMessaging();
+
+    // Enable Firebase Auto Init iff either notifications or analytics is enabled
+    var firebase_auto_init =
+        (await getNotificationsEnabled()) || (await getAnalyticsEnabled());
+    await firebaseMessaging.setAutoInitEnabled(firebase_auto_init);
```
brunobowden, line 112: Remove the variable above and inline in. I think the indirection detracts from readability rather than adds to it.

```suggestion
    await FirebaseMessaging().setAutoInitEnabled(firebase_auto_init);
```

brunobowden (approved):

brunobowden:
@epicfaace - please merge this as soon as you can. There's one slight improvement I suggested to inline a variable but really we just need to get it landed so we can make progress on other items. Thanks.

epicfaace:
Ok, will inline it later then

Issue #1625, opened 2020-10-17
Logs Terraform


brunobowden:
## Summary

I suspect that the logs are automatically created by the App Engine instance from Terraform. Nonetheless, it should explicitly set the following to ensure greater privacy:
1. 30 days retention
1. location inherited from project, comment should point to default location to reassure people of privacy

Documentation:
- https://www.terraform.io/docs/providers/google/r/logging_project_sink.html

## Checklist:

- [x] Searched the [existing issues](https://github.com/WorldHealthOrganization/app/issues) to ensure you are not creating a duplicate.
- [x] Followed the [Contributor Guidelines](https://github.com/WorldHealthOrganization/app/blob/master/docs/CONTRIBUTING.md).

PR #1738, opened 2020-11-14
Logging retention of 30 days made explicit

matthewblain:
Adds a new section to the MyHealth Service Terraform configuration: Logging bucket configuration.
This has Terraform maintain the logging retention on the default bucket at 30 days.

Fixes issue #1625 .

- [ ] iOS Simulator
- [ ] iOS Device
- [ ] Android Simulator
- [ ] Android Device
- [x] other, please describe

Ran Terraform Apply several times. Also tested with different values to values to manually verify that this setting applies.

## Checklist:

### DIFF
--- a/server/terraform/modules/myhealth/main.tf
+++ b/server/terraform/modules/myhealth/main.tf
@@ -82,8 +82,6 @@ resource "google_app_engine_application" "gae" {
   #}
 }
 
-# Logging storage
-# The _Default logging bucket is created automatically and set to 30 
 # Logging Storage 30 Days Retention
 # Ensures that logging records are retained for 30 days only
 # Matches the standard retention period of the _Default logging bucket: 
--- a/server/terraform/modules/myhealth/main.tf
+++ b/server/terraform/modules/myhealth/main.tf
@@ -82,6 +82,15 @@ resource "google_app_engine_application" "gae" {
   #}
 }
 
+# Logging storage
+# The _Default logging bucket is created automatically and set to 30 
+# days retention; this resource ensures it is kept at that length.
+resource "google_logging_project_bucket_config" "default" {
+    project    = var.project_id # google_project.project.name
+    location  = "global"
+    retention_days = 30
+    bucket_id = "_Default"
+}
 
 # Load Balancer
 module "lb" {
--- a/server/terraform/modules/myhealth/main.tf
+++ b/server/terraform/modules/myhealth/main.tf
@@ -86,10 +86,10 @@ resource "google_app_engine_application" "gae" {
 # The _Default logging bucket is created automatically and set to 30 
 # days retention; this resource ensures it is kept at that length.
 resource "google_logging_project_bucket_config" "default" {
-    project    = var.project_id # google_project.project.name
-    location  = "global"
-    retention_days = 30
-    bucket_id = "_Default"
+  project        = var.project_id # google_project.project.name
+  location       = "global"
+  retention_days = 30
+  bucket_id      = "_Default"
 }
 
 # Load Balancer


server/terraform/modules/myhealth/main.tf
```diff
@@ -82,6 +82,15 @@ resource "google_app_engine_application" "gae" {
   #}
 }
 
+# Logging storage
+# The _Default logging bucket is created automatically and set to 30 
+# days retention; this resource ensures it is kept at that length.
+resource "google_logging_project_bucket_config" "default" {
+    project    = var.project_id # google_project.project.name
+    location  = "global"
```
brunobowden, line 90: WHO Policy is to use Switzerland wherever possible, so it should inherit the same location as used throughout the rest of the project:
```suggestion
    location  = var.region
```

server/terraform/modules/myhealth/main.tf
```diff
@@ -82,6 +82,15 @@ resource "google_app_engine_application" "gae" {
   #}
 }
 
+# Logging storage
+# The _Default logging bucket is created automatically and set to 30 
+# days retention; this resource ensures it is kept at that length.
```
brunobowden, line 87:
```suggestion
# Logging Storage 30 Days Retention
# Ensures that logging records are retained for 30 days only
# Matches the standard retention period of the _Default logging bucket: 
# https://cloud.google.com/logging/docs/buckets#overview
```

brunobowden (approved):

brunobowden:
@matthewblain - I've removed the "prod:" prefix in the title as this will apply to all projects

server/terraform/modules/myhealth/main.tf
matthewblain: This is not an option for the _Default bucket. It does look like this is possible with additional configuration: https://cloud.google.com/logging/docs/regionalized-logs

matthewblain:
See note about region.
BTW, I used "prod:" as a prefix for Production, since there's no similar option in the list of types in the CONTRIBUTING doc. But it looks like recent commits don't have the type prefix?

brunobowden:
@matthewblain - regionalized logs is very important, so please incorporate that in to this PR. I had wondered if it would be automatically located in the same region anyway... but we need to be explicit about that.

brunobowden:
@matthewblain - I'll update the CONTRIBUTING guidelines as they're a little heavyweight and I haven't seen anyone following them.

brunobowden (approved):
Add comment about working on data privacy and then it's good to merge.

server/terraform/modules/myhealth/main.tf
```diff
@@ -82,6 +82,16 @@ resource "google_app_engine_application" "gae" {
   #}
 }
 
+# Logging Storage 30 Days Retention
+# Ensures that logging records are retained for 30 days only
+# Matches the standard retention period of the _Default logging bucket: 
+# https://cloud.google.com/logging/docs/buckets#overview
+resource "google_logging_project_bucket_config" "default" {
+  project        = var.project_id # google_project.project.name
+  location       = "global"
```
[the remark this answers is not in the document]
brunobowden: Better yet, have a link to the issue:
```suggestion
  # TODO: working on data location for privacy
  # https://github.com/WorldHealthOrganization/app/issues/1742
  location       = "global"
```

server/terraform/modules/myhealth/main.tf
brunobowden: @matthewblain - use `europe-west1` (Belgium) for now. We'll switch to Switzerland once it's available.

matthewblain:
I plan to abandon this in favor of pull request 1753 , regionalized config. It also includes the retention policy.

brunobowden:
Replaced by #1753

Issue #649, opened 2020-03-30
Network Security: TLS 1.2 min, SSL and ingress restrictions Terraform


myiremark:
Adding the network security requires allocating a global static IP address and load balancer. Once added, the load balancer allows you to apply the TLS and SSL restrictions. This is configured for the WHO GCP production server but needs to be added to Terraform.

Google Setup:
1. Reserve IP Address [Ref](https://cloud.google.com/load-balancing/docs/negs/setting-up-serverless-negs#ip-address)
   - Network Service Tier: Premium
   - IP Version: IPv4
   - Type: Global
1. Load Balancer [Ref](https://cloud.google.com/load-balancing/docs/negs/setting-up-serverless-negs#creating_the_load_balancer)
   - Type: Layer 7 HTTPS
      - Backend:
         - Backend: Serverless network endpoint group
         - Protocol: HTTPS
         - New Serverless network endpoint group
            - Service: App Engine
         - Enable Cloud CDN:
            - Select "Use origin settings based on Cache-Control headers" to reduce chance of mistaken caching of private data
      - Frontend:
         - Protocol: HTTPS
         - IP: as reserved above
         - TLS: minimum TLS 1.2
         - Profile: modern (incase we support a web app in the future)
1. http => https redirect
   - https://cloud.google.com/load-balancing/docs/https/setting-up-http-https-redirect#partial-http-lb
1. Cloud Armor on the same backend
   - No configuration, so I presume it's applying standard defaults
1. App Engine:
   - App Engine Firewall [ref](https://console.cloud.google.com/appengine/firewall?project=who-myhealth-staging)
   - Ingress limited to load balancer only and disable custom domain [Ref](https://cloud.google.com/appengine/docs/standard/python/application-security#ingress_controls). Example as command line but needs to be implemented in Terraform:
```
gcloud beta app services update default --ingress internal-and-cloud-load-balancing
```
After setup, we'll need to manually update the DNS records with the static IP addresses.

## Problem
TLS 1.0 and 1.1 are enabled on https://staging.whocoronavirus.org/. This is returning a B rating from SSL labs:
https://www.ssllabs.com/ssltest/analyze.html?d=staging.whocoronavirus.org&latest

## Testing

Need to check SSL Lab status on all the domains: staging and production

**Please follow this checklist. Please check each appropriate box (put an 'x' or check it after creating the issue).**
[...]

PR #1756, opened 2020-11-18
Require Load Balancer for all App Engine traffic

brunobowden:
- App Engine ingress set to `internal-and-cloud-load-balancing`
- Disables who-mh-hack.appspot.com domain
- Partial fix for #649

```
server/bin/deploy-server.sh who-mh-hack
```

## Checklist:

### DIFF
--- a/server/bin/deploy-server.sh
+++ b/server/bin/deploy-server.sh
@@ -8,5 +8,9 @@ PROJECT=$1
 ../tools/verify-server.sh
 
 ./gradlew build
-gcloud beta app deploy --quiet --project=$PROJECT appengine/build/war
-gcloud beta app deploy --quiet --project=$PROJECT appengine/build/war/WEB-INF/cron.yaml
+gcloud beta app --quiet --project=$PROJECT deploy appengine/build/war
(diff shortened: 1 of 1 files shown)


matthewblain (approved):
I'm surprised that this does not seem possible in Terraform. Would you like (or like me) to file a bug? The app_engine_split_traffic resource is half way there. Using gcloud from terraform (as in PR 1752) feels like a hack. In fact, I'd think this entire script should be doable in Terraform, without any 'gcloud' commands, though I don't know if Terraform yet supports all the things required.

matthewblain:
Also can you title this 'Require Load Balancer....'
(The LB was configured elsewhere.)

brunobowden:
@matthewblain - please file a bug. I agree that it should all be under Terraform. Ideally you'd specify a tag or commit sha. Then as soon as it's merges, Terraform would run, do the pinned build, update the server binaries and the server state at the same time.

For now though, we need to focus on getting something out earlier by which we can do more good.

@yukuairoy - need your approval too

yukuairoy (approved):

brunobowden:
@matthewblain - the idea is to put all the infrastructure under source control. Not just the configuration but also the deployment. So production and staging servers would all update automatically based on tag or sha which itself was under source control.

PR #1762, opened 2020-11-19
Create placeholder IP blocking security policy

### LINKED ISSUE #1758: Cloud Armor in Terraform: compute_security_policy
## Summary

Setup Cloud Armor with sensible values. Likely this will include some defaults and a configured but not used IP block list. The configured but not used list is so that we can test the system... and if we need to deploy it in a rush then it will be easy to change data rather than write code.

It's probably the DDOS protection that's most helpful but I'm not sure how to configure that in Terraform.

https://www.terraform.io/docs/providers/google/r/compute_security_policy.html
https://www.terraform.io/docs/providers/google/r/compute_backend_service.html#security_policy
https://cloud.google.com/armor/docs/rule-tuning#preconfigured_rules

## Checklist

matthewblain:
Create a placeholder Google Cloud Armor security policy blocking IP addresses.
Uses a loopback placeholder with a CIDR and IP example.

To edit interactively, view https://console.cloud.google.com/net-security/securitypolicies/lb-security-policy/rules/1000/edit

Fixes #1758

## Screenshots

<details>
</details>

- [ ] iOS Simulator
- [ ] iOS Device
- [ ] Android Simulator
- [ ] Android Device
- [x] other, please describe

Configured my test project with terraform apply
Manually changed settings in console, and tested that terraform apply apply un-did the changes.
Manually changed settings in console, made same settings in terraform config, and tested that apply did nothing.

## Checklist:

### DIFF
--- a/server/terraform/modules/myhealth/main.tf
+++ b/server/terraform/modules/myhealth/main.tf
@@ -212,8 +212,7 @@ resource "google_compute_security_policy" "policy" {
     description = "Deny access to specified IPs"
   }
 
-  # Every security policy must re-iterate the default rule at maxint32.
-  # The 'action' may be changed but not the other properties.
+  # The default rule at maxint32. Should not need to be changed.
   rule {
     action   = "allow"
     priority = "2147483647"
--- a/server/terraform/modules/myhealth/main.tf
+++ b/server/terraform/modules/myhealth/main.tf
@@ -226,7 +226,7 @@ resource "google_compute_security_policy" "policy" {
     description = "default rule"
   }
 
-  # In case  this resource is renamed in Terraform, the lifecycle must be:
+  # In case this resource is renamed in Terraform, the lifecycle must be:
   #  1: Create new policy 
   #  2: Configure backend to reference it
   #  3: Delete old policy.
--- a/server/terraform/modules/myhealth/main.tf
+++ b/server/terraform/modules/myhealth/main.tf
@@ -190,6 +190,52 @@ resource "google_compute_region_network_endpoint_group" "neg" {
   depends_on = [google_project_service.service]
 }
 
+
+# Google Cloud Armor security policy.
+resource "google_compute_security_policy" "policy" {
+  name = "lb-security-policy"
+
+  # IP Block list. 
+  rule {
+    action   = "deny(403)"
+    priority = "1000"
+    match {
+      versioned_expr = "SRC_IPS_V1"
+      config {
+        # Using an unused portion of the loopback block as a placeholder.
+        # Remove it once a real set of IP Ranges are present.
+        # Up to ten IP ranges can be created per rule.
+        src_ip_ranges = ["127.0.0.255/32","127.0.0.254"]
+      }
+    }
+    description = "Deny access to specified IPs"
+  }
+
+  rule {
+    action   = "allow"
+    priority = "2147483647"
+    match {
+      versioned_expr = "SRC_IPS_V1"
+      config {
+        src_ip_ranges = ["*"]
+      }
+    }
+    description = "default rule"
+  }
+
+  # In case  this resource is renamed in Terraform, the lifecycle must be:
+  #  1: Create new policy 
+  #  2: Configure backend to reference it
+  #  3: Delete old policy.
+  # To accomplish this, it is set to create_before_destroy, 
+  # then running apply twice completes the operation.
+  lifecycle {
+    create_before_destroy = true
+  }
+
+}
+
+
 resource "google_compute_backend_service" "backend" {
   name                            = "${var.project_id}-backend-appengine"
   load_balancing_scheme           = "EXTERNAL"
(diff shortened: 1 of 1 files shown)


server/terraform/modules/myhealth/main.tf
```diff
@@ -190,6 +190,52 @@ resource "google_compute_region_network_endpoint_group" "neg" {
...
+        # Up to ten IP ranges can be created per rule.
+        src_ip_ranges = ["127.0.0.255/32","127.0.0.254"]
+      }
+    }
+    description = "Deny access to specified IPs"
+  }
+
+  rule {
+    action   = "allow"
+    priority = "2147483647"
```
brunobowden, line 216:
```suggestion
    priority = "2147483647"   # max 32 bit signed int
```

server/terraform/modules/myhealth/main.tf
```diff
@@ -190,6 +190,52 @@ resource "google_compute_region_network_endpoint_group" "neg" {
...
+    match {
+      versioned_expr = "SRC_IPS_V1"
+      config {
+        src_ip_ranges = ["*"]
+      }
+    }
+    description = "default rule"
+  }
+
+  # In case  this resource is renamed in Terraform, the lifecycle must be:
```
brunobowden, line 226: Double space should be single space for:
```
case  this
```

server/terraform/modules/myhealth/main.tf
```diff
@@ -190,6 +190,52 @@ resource "google_compute_region_network_endpoint_group" "neg" {
...
+  #  1: Create new policy 
+  #  2: Configure backend to reference it
+  #  3: Delete old policy.
+  # To accomplish this, it is set to create_before_destroy, 
+  # then running apply twice completes the operation.
```
brunobowden, line 231: Is the logic reversed here? Comment said "if resource is renamed, then lifecycle must be..."

I'm expecting "can't update in place, therefore need to rename when updating", I did this with the random suffix on the ssl certificate name:

https://github.com/WorldHealthOrganization/app/blob/master/server/terraform/modules/http-load-balancer/main.tf#L71-L84

brunobowden:
Not sure why the precommit checks aren't running but here's the command to format Terraform:
```
cd server/terraform
terraform fmt -recursive .
```

server/terraform/modules/myhealth/main.tf
matthewblain: In what sense?
I could delete this, as no one really 'should' be renaming this resource from name = "lb-security-policy" to something else. And this is a bit of a distraction.

But what happens is that google_compute_backend_service.backend.security_policy references google_compute_security_policy.policy by name. So if you destroy nameOld and create nameNew, the steps need to be the three outlined above. Perhaps terraform has some way to describe this set of operations--it seems to know that google_compute_security_policy.policy needs to be ready to go for google_compute_backend_service.backend to be possible, but does not know that if the value is *changing* that it needs to be unreferenced first.
(Note: this could also be solved by having terraform unattach the 'old' security policy, then reattach it. That would mean updating google_compute_backend_service.backend twice, including having it be in a temporary state with no policy at all.)

matthewblain:
I have 
* Re-run terraform fmt. (No, I don't know why the precommit is erratic on my machine. It is running some items!)
* Updated the comments to be hopefully more useful to a future user of this block.

server/terraform/modules/myhealth/main.tf
brunobowden: Ok, not much of an issue then

server/terraform/modules/myhealth/main.tf
```diff
@@ -190,6 +190,55 @@ resource "google_compute_region_network_endpoint_group" "neg" {
...
+        # Using an unused portion of the loopback block as a placeholder.
+        # Remove it once a real set of IP Ranges are present.
+        src_ip_ranges = ["127.0.0.255/32", "127.0.0.254"]
+      }
+    }
+    description = "Deny access to specified IPs"
+  }
+
+  # Every security policy must re-iterate the default rule at maxint32.
+  # The 'action' may be changed but not the other properties.
```
brunobowden, line 216: Should we have the policy that this rule is never changed? I'm thinking about avoiding mistakes in the future if someone doesn't understand what they're doing. Main point is that only the `src_ip_ranges` should be edited in the block list.

brunobowden (approved):

yukuairoy (approved):

server/terraform/modules/myhealth/main.tf
matthewblain: I could change the comment to effectively say do not touch, which is the right setting for this project.

Medium term, I think the entire IP blocking config will move into its own config. It is in a sense 'individual remote client' level data (a list of probable malicious user addresses) and not 'production system' level data. So running a complete terraform apply will... cause issues if the two lists ever get out of sync. (E.g. need to update server and accidentally clobber block list; or need to update block list but not bring server config to latest.)

But for now it's probably simplest to keep it here.

PR #1770, opened 2020-11-21
Architecture diagram is not current

advayDev1:
Remove my lucidchart - my subscription is no longer alive and I of course am not updating the diagram at that URL.

Note also that the architecture diagram has its own special license (still under a partial MIT license).  I'm happy to relicense if the other authors are as well.

Some missing things from the diagram include but are not limited to:
- The TerraForm work that has been done recently and how those tools have admin access to production
- The storage of a portion of the server logs in Belgium (while others appear to be global)
- The location (presumably in the US) of the members of the current team who will access the logs and user databases
- The load balancers/reverse proxies that may live in front of the app engine
- and other things I may not be aware of (I only looked cursorily)

### DIFF
--- a/docs/devdesign/architecture.md
+++ b/docs/devdesign/architecture.md
@@ -1,4 +1,4 @@
 ## System Architecture
 
-- We keep our live diagram in LucidChart: https://www.lucidchart.com/documents/view/3ce4e176-551f-4e71-9ec5-1c2e38071dea/0_0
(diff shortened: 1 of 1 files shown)


yukuairoy (approved):

brunobowden (approved):
Thanks @advayDev1. I noticed that a while back but didn't get around to fixing it.

advayDev1:
Thx - can you merge?  I ofc can't,

Issue #1624, opened 2020-10-17
Cloud Storage (static content) Terraform


brunobowden:
## Summary

Cloud Storage static content bucket needs to be created and managed by Terraform. This requires configuring:
1. Public Bucket
1. Global CDN

The staging and hack servers should be automatically pushed on merge by adapting this existing automatic push mechanism:
- https://github.com/WorldHealthOrganization/app/blob/master/.github/workflows/static-content.yaml

## Checklist:

- [x] Searched the [existing issues](https://github.com/WorldHealthOrganization/app/issues) to ensure you are not creating a duplicate.
- [x] Followed the [Contributor Guidelines](https://github.com/WorldHealthOrganization/app/blob/master/docs/CONTRIBUTING.md).

brunobowden:
3rd requirement moved to #1768. Making this complete

3. Updating clients with urls for accessing content based on staging / hack / prod instances
- https://www.terraform.io/docs/providers/google/r/storage_bucket.html

PR #1771, opened 2020-11-22
Terraform cloud bucket configuration

matthewblain:
## Screenshots

<details>
</details>

- [ ] iOS Simulator
- [ ] iOS Device
- [ ] Android Simulator
- [ ] Android Device
- [x] other, please describe

Run terraform apply and look at the results in the Cloud Console

Manually added a file to the bucket, and verified that it could be loaded. Note: Prefix of filename in bucket must match prefix of filename in load balancer config. E.g. https;//hostname/content/example.txt must be stored as /content/hello.txt in the bucket.

This is steps one and two of issue #1624 .

## Checklist:

### DIFF
--- a/server/terraform/modules/myhealth/main.tf
+++ b/server/terraform/modules/myhealth/main.tf
@@ -35,6 +35,10 @@ provider "random" {
   version = "~> 3.0.0"
 }
 
+locals {
+  content_bucket_id = (var.content_bucket_id != "" ?
+  var.content_bucket_id : "who-app-devel-${var.project_id}")
+}
 
 # Google Project
 resource "google_project" "project" {
@@ -171,6 +175,10 @@ resource "google_compute_url_map" "urlmap" {
   path_matcher {
     name            = "all"
     default_service = google_compute_backend_service.backend.self_link
+    path_rule {
+      paths   = ["/content"]
+      service = google_compute_backend_bucket.content.id
+    }
 
   }
   depends_on = [google_project.project]
@@ -207,4 +215,58 @@ resource "google_compute_backend_service" "backend" {
   # TODO: figure out what health checks are possible
   health_checks = null
   depends_on    = [google_project.project]
+}
+
+
+# Google Cloud bucket for static content assets.
+resource "google_storage_bucket" "content" {
+  name          = local.content_bucket_id
+  location      = var.bucket_location
+  force_destroy = false ### ???? The documentation on this is atrocious. I believe if this is false it will fail on destruction if there is data in the bucket?
+
+  # There should be no need for per-object ACLs with this bucket.
+  uniform_bucket_level_access = true
+
+  # Do we need to set the lifecycle rule or object versioning? I expect not, or with a single version / short lifespan?
+  # logging - is the default to use the default sink or null or something else?
+}
+
+
+## Should we use google_storage_bucket_iam_binding or google_iam_policy? 
+## google_iam_policy is really dangerous... if you 'destroy' it,  it does not return to the default, but to empty!
+
+#data "google_iam_policy" "public-viewer" {
+#  binding {
+#    role = "roles/storage.objectViewer"
+#    members = [
+#      "allUsers",
+#    ]
+#  }
+#  
+#  binding {
+#    role = "roles/storage.admin"
+#    members = [
+#      "projectOwner:${var.project_id}",
+#    ]
+#  }
+#}
+#
+#resource "google_storage_bucket_iam_policy" "content" {
+#  bucket = google_storage_bucket.content.name
+#  policy_data = data.google_iam_policy.public-viewer.policy_data
+#}
+
+resource "google_storage_bucket_iam_binding" "binding" {
+  bucket = google_storage_bucket.content.name
+  role   = "roles/storage.objectViewer"
+  members = [
+    "allUsers",
+  ]
+}
+
+# Routing
+resource "google_compute_backend_bucket" "content" {
+  name        = "static-content-backend-bucket"
+  bucket_name = google_storage_bucket.content.name
+  enable_cdn  = true
 }
--- a/server/terraform/modules/myhealth/main.tf
+++ b/server/terraform/modules/myhealth/main.tf
@@ -242,7 +242,6 @@ resource "google_compute_security_policy" "policy" {
   lifecycle {
     create_before_destroy = true
   }
-
 }
 
 
--- a/server/terraform/modules/myhealth/main.tf
+++ b/server/terraform/modules/myhealth/main.tf
@@ -35,11 +35,6 @@ provider "random" {
   version = "~> 3.0.0"
 }
 
-locals {
-  content_bucket_id = (var.content_bucket_id != "" ?
-  var.content_bucket_id : "who-app-devel-${var.project_id}")
-}
-
 # Google Project
 resource "google_project" "project" {
   provider        = google-beta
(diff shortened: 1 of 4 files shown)


server/terraform/modules/myhealth/main.tf
```diff
@@ -171,6 +175,10 @@ resource "google_compute_url_map" "urlmap" {
   path_matcher {
     name            = "all"
     default_service = google_compute_backend_service.backend.self_link
+    path_rule {
+      paths   = ["/content/*"]
```
brunobowden, line 179: So `covid19app.who.int/content` will be the production url? We don't have a website yet so a separate subdomain for static content isn't as much of a concern.

The app currently fetches the content from a Google Storage managed url. I don't have a clear opinion on which is the better approach except that we should probably use only one of them.
https://storage.googleapis.com/who-myhealth-staging-static-content-01

From:
https://github.com/WorldHealthOrganization/app/blob/master/client/lib/api/endpoints.dart#L8-L9

server/terraform/modules/myhealth/main.tf
```diff
@@ -257,3 +264,57 @@ resource "google_compute_backend_service" "backend" {
   health_checks = null
   depends_on    = [google_project.project]
 }
+
+
+# Google Cloud bucket for static content assets.
+resource "google_storage_bucket" "content" {
+  name          = local.content_bucket_id
+  location      = var.bucket_location
+  force_destroy = false ### ???? The documentation on this is atrocious. I believe if this is false it will fail on destruction if there is data in the bucket?
```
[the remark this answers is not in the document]
brunobowden: Put it at the end of the resource but this acts to prevent destruction of the resource which is what I think you want. Many other resources can be deleted and recreated without as much of an issue. If someone removes the `lifecyle` section, e.g. by deleting the whole resource, then it won't protect it.... but at least it acts as a prompt for accidental changes or a note in the code that someone should be extra careful.
```suggestion
  lifecycle {
    prevent_destroy = true
  }
```

server/terraform/modules/myhealth/main.tf
```diff
@@ -257,3 +264,57 @@ resource "google_compute_backend_service" "backend" {
   health_checks = null
   depends_on    = [google_project.project]
 }
+
+
+# Google Cloud bucket for static content assets.
+resource "google_storage_bucket" "content" {
+  name          = local.content_bucket_id
+  location      = var.bucket_location
```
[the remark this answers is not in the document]
matthewblain: OK; the fact that there's a CDN in front of this makes that totally sensible.

server/terraform/modules/myhealth/variables.tf
```diff
@@ -10,6 +10,23 @@ variable "domain" {
   type        = string
 }
 
+variable "content_bucket_id" {
+  description = "Name of Cloud Storage bucket to provision for static content."
+  type        = string
+  # Default will generate a name using the project ID.
+  default = ""
+}
```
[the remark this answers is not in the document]
matthewblain: Definitely better if possible. That presumably means using entirely new buckets and deleting the current ones once all client apps have started using the load balancer to access them. What pattern would you like to use? It's in a flat global namespace though it can sometimes use appropriately owned DNS domain names. I'll propose something like 

who-myhealth-$projectid-static-content

server/terraform/modules/myhealth/main.tf
matthewblain: Yes, that is what this is configuring here. It would allow access via either path (storage.googleapis or who.int). 
Not sure of best practices for separate subdomains (rpc vs static). My reading of the bug was to route everything through the LB.  And my assumption is that this content is not user created so I can't see any major issues with sharing.
Certainly if there is a desire to front/move everything on a different system (e.g. azure) using a custom domain seems better than storage.googleapis...

server/terraform/modules/myhealth/main.tf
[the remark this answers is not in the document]
matthewblain: will do both. i believe force_destroy = false means it will fail on destroy if the bucket is non-empty; force_destroy=true means it'll zap the contents. but the documentation has so many double negatives and clauses with uncertain self-references.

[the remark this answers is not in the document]
matthewblain: note that the prevent_destroy will block if the bucket is renamed (like i am doing now)... probably a good thing, certainly can be worked around and remind the operator they're doing something very strange.

matthewblain:
Updated; see comments in code with long blocks of ### .

server/terraform/modules/myhealth/variables.tf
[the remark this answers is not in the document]
brunobowden: The `project_id` already includes `who-mh-xxx`, so that's all you need. Here's some example from `terraform/modules/myhealth/main.tf`:
```
  name = "${var.project_id}-lb"
  name = "${var.project_id}-url-map"
  name = "${var.project_id}-neg-appengine"
```

server/terraform/modules/myhealth/main.tf
brunobowden: The system is pretty heavily tied to GCP already.

The benefit of the storage url is that everything is managed for you and probably less likely to fail. Slight downside is that the domain isn't associated with WHO so it could lead to different filtering. I didn't mean for the bug to take a stance on using the LB or not. Perhaps that best so everything goes through the same domain.

server/terraform/modules/myhealth/main.tf
```diff
@@ -257,3 +259,62 @@ resource "google_compute_backend_service" "backend" {
   health_checks = null
   depends_on    = [google_project.project]
 }
+
+
+# Google Cloud bucket for static content assets.
+resource "google_storage_bucket" "content" {
+  name     = "${var.project_id}-static-content" #################### Should a prefix be added to make uniqueness more likley?
```
brunobowden, line 266: There's already uniqueness in project ids.... so I'd think it unlikely that someone would coincidentally use the same prefix. There's been no collisions so far... so we'll just worry about it if and when it happens.

It seems like the names have to be globally unique:https://cloud.google.com/apis/design/resource_names

server/terraform/modules/myhealth/main.tf
```diff
@@ -257,3 +259,62 @@ resource "google_compute_backend_service" "backend" {
...
+# Google Cloud bucket for static content assets.
+resource "google_storage_bucket" "content" {
+  name     = "${var.project_id}-static-content" #################### Should a prefix be added to make uniqueness more likley? 
+  location = var.region
+
+  # There should be no need for per-object ACLs with this bucket.
+  uniform_bucket_level_access = true
+
+  ############################################################
+  # Do we need to set the lifecycle rule or object versioning? I expect not, or with a single version / short lifespan?
```
brunobowden, line 273: We probably don't need to add `prevent_destroy` but it may prevent an accident or two.

server/terraform/modules/myhealth/main.tf
matthewblain: After this change, both options become (and remain) available. I'm fine either way and can remove this or leave it in.

brunobowden: Leave it in for now and let's discuss with the Google Cloud expert they introduced us to

brunobowden (approved):
@matthewblain - still a few issues for you to address but otherwise looking good. Please merge in when you're ready.

server/terraform/modules/myhealth/main.tf
matthewblain: Right, there's multiple namespaces.
$project_id = globally unique. Flat namespace. (Sort of; there are or were a certain class of scoped IDs but I think that's gone?) Not reusable. One of two true immutable indentifiers for the Google Cloud project. Exposed in a few places, including the appengine URL (foo.appspot.com, though for reasons I don't understand they added foo.someregionalgibberish.appspot.com) .
Various resources which are scoped to the project itself, e.g. most (but not all) of the things using projectid.something in this config.
Bucket ID. Globally unique. Both flat and domain owned namespace. Names can be re-used shortly after they are purged. (Note: every project gets a bucket named $projectid.appspot.com, which is guaranteed to be available due to the properties of the projectid described above).

Anyway, yes using $projectid-static-content has a very high chance of being available; if it ever breaks this config will just need more wackyness to manage it.

In any case, the app code needs to be updated to use this new bucket URL, be it using this name or the load balancer which hides the bucket name.

server/terraform/modules/myhealth/main.tf
matthewblain: Ah, I was actually talking about the policy of the bucket, not the terraform management.

https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/storage_bucket#lifecycle_rule

I think it's OK to leave this the default, as the contents of this bucket should all be pushed from an external source, the bucket is a distribution mechanism not the source of the data.

matthewblain:
Made some small tweaks

brunobowden (approved):
@matthewblain - thanks

The can be done in another PR. The deployment scripts should be refactored to extend the “deploy-server.ah” script..... parameterized used by project id: https://github.com/WorldHealthOrganization/app/blob/master/server/bin/deploy-server.sh

PR #1777, opened 2020-11-24
Privacy as Code policy

brunobowden:
- Production server changes must go through review first
- Other servers should go through review first as well
- Exception for emergencies

## Checklist:

### DIFF
--- a/server/terraform/README.md
+++ b/server/terraform/README.md
@@ -1,9 +1,61 @@
-## Terraform
+# Server Setup
+
+Documentation on settings up the servers from the Terraform to manually configured
+settings.
+
+## Policies
+
+The team has adopted a number of policies to improve transparency, security and
+auditability of the project. The main principle is:
+
+**All Production server changes should go through review before being applied
+and should be public.**
+
+This review includes manual changes to settings, e.g. Analytics. All of which
+are documented under source control. They should be code reviewed and merged
+before being applied in production. This page including the [Terraform
+Setup](https://github.com/WorldHealthOrganization/app/blob/master/server/terraform/README.md#Terraform-Setup)
+and [Manual Setup](https://github.com/WorldHealthOrganization/app/blob/master/server/terraform/README.md#Terraform-Setup)
+are the definitive record of the server setup. Other documents should be
+updated to match this.
+
+### Infrastructure as Code (IaC)
 
 Terraform is used for [Infrastructure as Code](https://en.wikipedia.org/wiki/Infrastructure_as_code) (Iac).
 Beyond the normal benefits of IaC, it serves essential roles for the WHO App
 for transparency and security. Team policy is to use IaC and in particular Terraform
 for as much configuration as possible. Particularly the WHO production servers.
+
+### Privacy as Code (PaC)
+
+The motivation is to improve the privacy and auditability of the server.
+This should allow for maximum transparency in the entire system from the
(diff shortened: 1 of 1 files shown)


advayDev1:
Some things to consider:
- Do you really want all server configs to be public - for example, even in non-emergency cases should permanent blockings of IP ranges in you DoS configs be in public source control (as you have the example in github right now)?
- Is it really possible to encode privacy rules in code at this stage of maturity of terraform?  In many cases it is the privacy rule you *can't* specify using terraform that will be the most interesting (for example, the various settings of Google Analytics data sharing, and the locations of storage/processing data that are not in Europe).  Those negative privacy assertions are not documented in your terraform (i.e. as comments), and cannot currently be implemented as normal Terraform code AFAIK.

You may be setting yourself up to create expectations from the other open source devs or the userbase for a level of transparency that is infeasible to meet.  And if you can't meet those expectations, consider whether wording things as strongly as you have it here (a so-called "policy" of Privacy as code) is misleading, and wether a weaker, more specific wording, would make clearer exactly which parts of the project's privacy rules can be specified in the github, and importantly the large portion which is not and cannot feasibly be.

brunobowden:
@advayDev1 - these are good comments. I've revised the policy to have more exceptions including those you've described. I've also documented in the top level Terraform about the configuration outside of Terraform. Please take a look.

advayDev1:
@brunobowden the concept that all privacy rules should be documented somewhere is a good one.  The Q is: is that actually the case?  For example, to do this correctly, you'll need to start re-updating the architecture diagram (I guess in Visio?) to document the location of various data storage and processing which has changed since September.

brunobowden:
@advayDev1 - added comment that this is the only document which should be considered "definitive" but that other documents "should" but updated to match. The ideas is that the code and manual setup instructions should completely describe the system. I think the exceptions for emergencies, intrusion and other security areas should suffice. We'll update them if needed.

Also got an approval from WHO for the policy too as this limits them also.

advayDev1:
@brunobowden my point is that a huge part of privacy can never be expressed in code.  For example: today, hypothetically, only you may be authorized by WHO to store and process user data in production and you're in the US.  But in a few weeks, let's say a new trusted volunteer is approved by WHO to have the same power as you, and they are in the UK.  How would any of this be expressed in Privacy as code, or in terraform?  Would each of your admins accounts and their locations be specified in terraform?  How would the github-perusing public know exactly who has access to their data, where, and when?  What about for third parties not managed by terraform, like Apple's services under APNS?

I just don't think it is possible to express privacy as code in general, and that's why we have contracts and DPAs and such which are often more definitive than code.

Since it isn't actually possible to express all of this in code, I think claiming that this gives complete transparency as stated in this commit, is incorrect.  Users can't get the full picture looking at this "definitive" document. In reality, it is only a small microcosm of the privacy stance of the project.

brunobowden:
@advayDev1 - you're right that there are limits. I'm trying to push as much as practically possible to do the best we can for privacy and transparency. All the servers that have been setup very carefully, following the documented steps very carefully.

I've posted another update that acknowledges that you can't have complete transparency and protecting the identity of the people with server access. Those people need to be prepared for all sorts of attacks but disclosing their identity would bring on more attacks. It is though entirely fit and proper that WHO approves those people individually and enforces appropriate restrictions on confidentiality, respecting the privacy policy and so forth.

matthewblain (approved):

PR #1784, opened 2020-11-27
Change static Cloud Storage bucket to multi-region location

### LINKED ISSUE #1783: Terraform config uses features not available in the required_version
## Summary

Terraform config has "required_version >= 0.12" - but in fact the config uses features that are only available in >= 0.13.0. Specifically, the "depends_on" attribute in the "lb" module definition.

https://github.com/hashicorp/terraform/blob/v0.13/CHANGELOG.md

terraform {
  required_version = ">= 0.12"
}

\# Load Balancer
module "lb" {
  source             = "../http-load-balancer"
  project_id         = var.project_id
  name               = "${var.project_id}-lb"
  url_map            = google_compute_url_map.urlmap.self_link
  create_dns_entries = var.create_dns_entry
  dns_record_ttl     = var.dns_record_ttl
  domain             = var.domain
  enable_http        = false
  **\# depends_on available from terraform 0.13
  depends_on         = [google_project_service.service]**
}

## Reproduction Steps

1. install Terraform 0.12.x
2. execute terraform init
3. See error:

> jt-dev% terraform init
> Initializing modules...
> - myhealth in ../modules/myhealth
> - myhealth.lb in ../modules/http-load-balancer
> There are some problems with the configuration, described below.
> 
> The Terraform configuration must be valid before initialization so that
> Terraform can determine which modules and providers need to be installed.
> 
> Error: Reserved argument name in module block
> 
>   on ../modules/myhealth/main.tf line 157, in module "lb":
>  157:   depends_on         = [google_project_service.service]
> 
> The name "depends_on" is reserved for use in a future version of Terraform.

## Expected Behavior

Terraform init completes without error.

## System

- MacOS
- Terraform v0.12.25

## Checklist:

matthewblain:
Updated location of Cloud Storage static content bucket to multi-region location.

This is done with a map to keep the number of variables small. Similar techinique could be done for the log storage bucket. Or conversely, they could both be a variable.

Note that moving the bucket location requires deleting/recreating the bucket and its contents.

Continues progress on issue WorldHealthOrganization#1624
Also fixes #1783 .

## Screenshots

<details>
</details>

- [ ] iOS Simulator
- [ ] iOS Device
- [ ] Android Simulator
- [ ] Android Device
- [x] other, please describe

Terraform apply in my dev cluster.

## Checklist:


server/terraform/modules/myhealth/main.tf
```diff
@@ -10,6 +10,19 @@ terraform {
...
+  # using var.region prefix; based on current list at:
+  # https://cloud.google.com/storage/docs/locations
+  storage_location_prefixes = {
+    "EUROPE" = "EU",
+    "US"     = "US",
+    "ASIA"   = "ASIA"
+  }
+  storage_bucket_location = lookup(local.storage_location_prefixes,
+  upper(split("-", var.region)[0]), var.region)
+}
```
brunobowden, line 24: The bucket still points to the old `var.region` location, so that needs to be updated in the terraform on line 282.

Google docs state that the location can't be changed except by destroying the bucket and creating a new bucket with the same name. There isn't a good way to do that in Terraform, so you may need to change the name in order for it to work on an existing deployment.
https://cloud.google.com/storage/docs/moving-buckets

server/terraform/modules/myhealth/main.tf
```diff
@@ -10,6 +10,19 @@ terraform {
...
+  # We want the storage bucket to be multi-region. Lookup the location for that
+  # using var.region prefix; based on current list at:
+  # https://cloud.google.com/storage/docs/locations
+  storage_location_prefixes = {
+    "EUROPE" = "EU",
+    "US"     = "US",
+    "ASIA"   = "ASIA"
+  }
+  storage_bucket_location = lookup(local.storage_location_prefixes,
+  upper(split("-", var.region)[0]), var.region)
```
brunobowden, line 23: Nice way to do this. We'll need to fix it up if we deployed in a region where there's no multi-region support, e.g. Australia... but mostly this will do the right thing.

brunobowden (changes requested):

server/terraform/modules/myhealth/main.tf
matthewblain: Exactly, if region is set to AUSTRALIA-SOUTHEAST1 then bucket location will also be AUSTRALIA-SOUTHEAST1.

server/terraform/modules/myhealth/main.tf
matthewblain: D'oh the merge got mangled and I didn't test after.
What happens if you move the locaiton is that Terraform will destory and re-create the bucket. GIven that this is not a common operation, the operator will first have to manually delete the contents of the bucket. Then run apply once or twice.  (Should be once but it messed up one of the dependencies when I did it, so I ran it twice and tweaked the tf file.)

server/terraform/modules/myhealth/main.tf
brunobowden: So it defaults to a single region if multi-region doesn’t exist? Document either way on that behaviour.

server/terraform/modules/myhealth/main.tf
brunobowden: We should add a `prevent_destroy` to the storage bucket as part of this PR.

brunobowden (approved):

server/terraform/modules/myhealth/main.tf
matthewblain: having had to destroy/recreate the bucket in my test project a few times while testing this, i'm happy with 
force_destroy = false. if the bucket has no content, destroying it basically harmless. if it has content, force_destroy will block its destruction.

brunobowden:
Looking good

PR #1786, opened 2020-11-28
Static Content deploy to use ProjectId

brunobowden:
- Rebuild scripts around the ProjectID
- Removed staging / prod specific scripts
- Support legacy mode for GitHub deploy of old staging server

- [ ] iOS Simulator
- [ ] iOS Device
- [ ] Android Simulator
- [ ] Android Device
- [x] other, please describe

Repeated runs of the scripts across different project and manually running expected GitHub workflow script.

## Checklist:

### DIFF
--- a/server/README.md
+++ b/server/README.md
@@ -31,7 +31,7 @@ Services:
 
 ## Curl Testing
 
-Against staging:
+#### App Engine:
 
 ```
 # App Engine
@@ -50,8 +50,13 @@ curl -i \
 	-X POST \
 	-d '{isoCountryCode: CH}' \
 	'https://whoapp.org/WhoService/putLocation'
+```
 
-# Static Content - served from Google Cloud Storage
+#### Static Content
+
+Served from Google Cloud Storage:
+
+```
 curl https://storage.googleapis.com/who-myhealth-staging-static-content-01/\
 content/bundles/protect_yourself.en_US.yaml
 ```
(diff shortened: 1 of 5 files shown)


tools/build-and-push-static-serving.sh
```diff
@@ -11,6 +12,17 @@ if [ "$DONT_BUILD" != "true" ]; then
...
+  BUCKET=$1
+  if ! [[ $BUCKET == who-* ]]; then
+    echo "Required argument: project id starting with 'who-' : $PROJECT"
+    exit 1
+  fi
+  # TODO: remove once .github/workflows/static-content.yaml is updated
+  if [[ $BUCKET == *static-content* ]]; then
+    echo "Deprecated argument for bucket name: $BUCKET"
+  else
+    BUCKET=$BUCKET-static-content
```
matthewblain, line 24: i assume this whole script will go away, but it should be possible to get the result from Terraform. Which sadly is quite difficult.

server/README.md
```diff
@@ -50,8 +50,13 @@ curl -i \
...
 	'https://whoapp.org/WhoService/putLocation'
+```
+
+#### Static Content
 
-# Static Content - served from Google Cloud Storage
+Served from Google Cloud Storage:
+
+```
 curl https://storage.googleapis.com/who-myhealth-staging-static-content-01/\
```
matthewblain, line 60: and the curl line below will change as well, but not yet.

matthewblain (approved):
i should be able to run this against my own copy of the project, right? will test soon... though not sure how i'll run sh. bash should work. and maybe WSL will run sh but I don't have the whole toolchain installed there.

brunobowden:
@matthewblain - I'm wondering if we should default to `/bin/bash` for all the scripts. After this change, we'll only have `/bin/sh` scripts, so it wouldn't be too onerous to switch them all over.

server/README.md
brunobowden: agreed

brunobowden:
@matthewblain - it could complain that your projectId doesn't start with `who-` but otherwise should work. Maybe we should have a convention of suffixing names liked projectId with the GitHub username to ensure uniqueness but that's over the top for now.

PR #1789, opened 2020-11-30
WIP Android Flavors

theswerd:
Android Build Flavors


brunobowden:
@theswerd - just use the prefix WIP for "Work in Progress", it's a shorter way of saying the same thing.

yukuairoy (approved):

client/lib/main.dart
```diff
@@ -50,7 +50,8 @@ void mainImpl({@required Map<String, WidgetBuilder> routes}) async {
     _packageInfo = await PackageInfo.fromPlatform();
   }
 
-  await Firebase.initializeApp();
+  var app = await Firebase.initializeApp();
+  print('PROJECT ID: ${app.options.projectId}');
```
brunobowden, line 54: Mention where the project is coming from and no need for ALL CAPS:
```
print('Firebase ProjectID: ${app.options.projectId}');
```

brunobowden (commented):
@theswerd - add something to the `client/README.md` explaining how to build the custom flavors. That same information should be repeated in the commit message to aid review and other developers.

client/android/app/build.gradle
```diff
@@ -51,6 +53,19 @@ android {
...
+    flavorDimensions app
+
+    productFlavors {
+        hack {
+            dimension app
+            versionNameSuffix "-hack"
+        }
+        prod {
+            dimension app
+            versionNameSuffix "-prod"
```
brunobowden, line 66: Add in staging flavor. This will leave the original staging (who-myhealth-staging) as the default.

@paulinar - do you know if the following is the right way to set a default build flavor? We want developers to default to the staging server... then they can get other servers with a specific flavor.

https://proandroiddev.com/android-set-a-default-build-type-flavor-a1cdcd1c15fa
```
productFlavors {
  ...
  staging {
    dimension server
    getIsDefault().set(true)
  }
}
```

client/android/app/build.gradle
```diff
@@ -16,6 +16,8 @@ if (flutterRoot == null) {
 def flutterVersionCode = localProperties.getProperty('flutter.versionCode')
 def flutterVersionName = localProperties.getProperty('flutter.versionName')
 
+def app = "app"
```
brunobowden, line 19: remove line

client/android/app/build.gradle
```diff
@@ -51,6 +53,19 @@ android {
             signingConfig signingConfigs.debug
         }
     }
+
+    flavorDimensions app
```
brunobowden, line 57: `app` is too ambiguous, so change to server instead:
https://developer.android.com/studio/build/build-variants#flavor-dimensions
```
flavorDimensions "server"
```

brunobowden (changes requested):
@theswerd - I can confirm this works for the crash reporting. Switching flavors also means that it doesn't have the problem of stale cache pointing to old Firebase server.

@paulinar - I'd appreciate your perspective on best practices here.

client/android/app/build.gradle
theswerd: kept line - updated it to be server instead

client/android/app/build.gradle
theswerd: Added these lines

client/android/app/build.gradle
```diff
@@ -51,6 +53,23 @@ android {
...
+            dimension server
+            versionNameSuffix '-hack'
+        }
+        prod {
+            dimension server
+            versionNameSuffix '-prod'
+        }
+        staging {
+            dimension server
+            getIsDefault().set(true)
```
brunobowden, line 70:
```suggestion
            # TODO: make the default once it's automatically pushed
            # For now the default is the old `who-myhealth-staging` json file
            # getIsDefault().set(true)
```

client/android/app/build.gradle
```diff
@@ -5,17 +5,19 @@ if (localPropertiesFile.exists()) {
...
 def flutterRoot = localProperties.getProperty('flutter.sdk')
 if (flutterRoot == null) {
-    throw new GradleException("Flutter SDK not found. Define location with flutter.sdk in the local.properties file.")
+    throw new GradleException('Flutter SDK not found. Define location with flutter.sdk in the local.properties file.')
 }
 
 def flutterVersionCode = localProperties.getProperty('flutter.versionCode')
 def flutterVersionName = localProperties.getProperty('flutter.versionName')
 
+def server = "server"
```
brunobowden, line 19: Move to immediately before `flavorDimensions`, it adds cognitive load to require someone to track down the earlier definition, no reason not to have them together.

brunobowden (changes requested):
Finally move the `google-services-staging.json` file to its own flavor. Leave the top level `google-services.json` file as is.

brunobowden:
Refactored in to #1798

Issue #1799, opened 2020-12-06
refreshCaseStats / cron job disabled by ingress restrictions


brunobowden:
## Summary

The `refreshCaseStats` cron job is failing due to the `--ingress internal-and-cloud-load-balancing` settings. The url request is at `/internal/cron/refreshCaseStats`, so I thought it would be allowed as internal request but apparently not. That also explains why there's no logs for the cron job as the request can't reach the server.

## Workaround

Disable the ingress settings:
```
gcloud beta app --quiet --project=who-mh-staging services update default --ingress all
# WORKS: cron job succeeds

gcloud beta app --quiet --project=who-mh-staging services update default --ingress internal-and-cloud-load-balancing
# FAILS: cron job fails as before
```

## Reproduction Steps

1. App Engine Cron Dashboard shows "Failed": https://console.cloud.google.com/appengine/cronjobs?project=who-mh-staging&organizationId=532343229286
2. Click on "View" shows zero logs: https://console.cloud.google.com/logs/query;query=protoPayload.taskQueueName%3D%22__cron%22?organizationId=532343229286&project=who-mh-staging

## Expected Behavior

Cron job to run successfully with the ingress restrictions in place.

## Screenshots

App Engine Cron dashboard shows "Failed":
https://console.cloud.google.com/appengine/cronjobs?project=who-mh-staging&organizationId=532343229286
<img width="897" alt="Screen Shot 2020-12-06 at 1 14 34 AM" src="https://user-images.githubusercontent.com/679339/101276263-738de800-3760-11eb-87d3-da87a0b65f23.png">

Clicking on "View" shows zero logs:
https://console.cloud.google.com/logs/query;query=protoPayload.taskQueueName%3D%22__cron%22?organizationId=532343229286&project=who-mh-staging
<img width="1171" alt="Screen Shot 2020-12-06 at 12 56 35 AM" src="https://user-images.githubusercontent.com/679339/101276246-52c59280-3760-11eb-8484-1a6695286c90.png">

## Checklist:

- [x] Searched the [existing issues](https://github.com/WorldHealthOrganization/app/issues) to ensure you are not creating a duplicate.
- [x] Followed the [Contributor Guidelines](https://github.com/WorldHealthOrganization/app/blob/master/docs/CONTRIBUTING.md).

matthewblain:
Yes, this seems to be the case. Put together a simple test which makes it easy for anyone to verify. 
https://github.com/matthewblain/appengine_cron_test

PR #1800, opened 2020-12-06
Endpoints changed based on build flavor

### LINKED ISSUE #1781: Configure backend and static content url based on build flavor
## Summary

The backend and static content urls need to be swapped out based on build flavor. Ideally this could be done entirely programmatically, so specifying any flavor would automatically pick up the correct server without needing additional work. At the same time, production at least will need special settings for the backend and the `covid19app.who.int` domain.

1. Change backend
1. Change static content url
1. Remove `BuildInfo.DEVELOPMENT_ONLY` and instead use `Endpoint.prod`
1. After it's merged in, update the github workflows (@brunobowden will do that part)

Here's how to get the project name at runtime from `main.dart`. You also want it to throw an error and crash if the `google-services.json` or `GoogleServices.plist` is missing or Firebase fails to initialize. There should be a `initSuccessful()` method that returns a bool but I haven't found it yet.....
```
  var fapp = await Firebase.initializeApp();
  // need to confirm this is the right test, check on both platforms that
  // it fails when no Google Services file is provided
  if (Firebase.apps.length == 0) {
    throw Exception('Firebase failed to load');
  }
```
Then further down the file:
```
            Provider(
              create: (_) => Endpoint.create(projectId);
```

The environment.dart output then changes according to the projectId. It's important that everything is const as that enforces compile time errors (final doesn't ensure this). You always want to move errors earlier, so from production into staging and QA, then in to development runtime error, then a compile time error and so on.
```
class Endpoints {
   static const defaultStaticContentUrl = 'https://storage.googleapis.com/{$projectId}';
   static const defaultServiceUrl = 'https://{$domain}.whocoronavirus.org';

   // Production is the only special case, everything else uses the defaults above
   static const prodServiceUrl = 'https://covid19app.who.int';
   static const prodProjectId = 'who-mh-prod';

   static const create(String projectId) {
     // pseudocode as I know this doesn't do the string substitution properly
     final staticContentUrl = defaultStaticContentUrl.substitute(projectId);
     final prod = projectId == prodProjectId;
     final serviceUrl = prod ? prodServiceUrl
       : defaultServiceUrl.substitute(projectId);

     return Endpoint(serviceUrl, staticContentUrl, prod);
   }
}

class Endpoint {
  final String service;
  final String staticContent;
  final bool prod;

  Endpoint({@required this.service, @required this.staticContent, @required this.prod});
}
```

## Checklist:

brunobowden:
- "No Privacy" alert based on "prod" server or not
- Updates App Engine and Content endpoints
- No more direct usage of Storage Bucket direct URL
- Fixes #1781

```
flutter run --flavor prod
```
Checked:
- "No Privacy..." message not shown but shown for other flavors
- Fetched content successfully
- Discovered another issue that's orthogonal: #1799

## Checklist:

### DIFF
--- a/client/lib/api/content/content_loading.dart
+++ b/client/lib/api/content/content_loading.dart
@@ -1,3 +1,4 @@
+import 'package:who_app/api/endpoints.dart';
 import 'package:who_app/api/who_service.dart';
 import 'package:flutter/services.dart' show rootBundle;
 import 'package:flutter/cupertino.dart';
@@ -13,8 +14,8 @@ class ContentService {
   static final Duration networkTimeout = Duration(seconds: 30);
   static final String baseAssetPath = 'assets/content_bundles'; // no trailing
 
-  ContentService({@required String endpoint})
-      : baseContentURL = '$endpoint/content/bundles';
+  ContentService({@required Endpoint endpoint})
+      : baseContentURL = '${endpoint.service}/content/bundles';
 
   /// Load a localized content bundle loaded preferentially from the network, falling back
   /// to a local asset.  If no bundle can be found with the specified name an exception is thrown.
--- a/client/lib/api/endpoints.dart
+++ b/client/lib/api/endpoints.dart
@@ -1,19 +1,29 @@
-import 'package:flutter/foundation.dart';
-
-class Endpoints {
-  static final String STAGING_SERVICE = 'https://staging.whocoronavirus.org';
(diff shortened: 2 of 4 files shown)


yukuairoy (approved):

theswerd (approved):
Lgtm

brunobowden:
@theswerd @yukuairoy - thanks

Issue #1524, opened 2020-09-25
putLocation and putDeviceToken in single request


advayDev1:
The mutator rpcs are prone to a race condition where the devicetoken  or location get overwritten because both rpcs happen in quick succession during onboarding.

The two mutator RPCs must, within their body, read the Client, update the Client, and save it within a transaction.

We've seen this in real life as well, the race condition is not merely theoretical.

brunobowden:
@crazybob - do you have any spare time that you can spend to work on this? I'm guessing it would be much faster for you than any of us. Any general pointers or even pseudocode would be helpful.

brunobowden:
FYI @yukuairoy

advayDev1:
IMO there's an easier solution that I also believe is simpler: just have a single rpc '/putClientInfo' to save all client info, with all attributes provided by the device each time.  The device always has the complete picture of its own configuration.

( Because the "read" of the Client happens in the ClientInterceptor and the "write" happens in the service implementation, I wasn't sure how to put the whole thing in a transaction (I'm not at all familiar with Objectify framework personally). )

brunobowden:
@advayDev1 - good idea. I'll work on that once I get the new build unless someone else beats me to it.

brunobowden:
@matthewblain - please take a look at this. Also see @advayDev1's suggestion of combining the two request in a single request to avoid the race condition. This sounds like a good approach to me:
https://github.com/WorldHealthOrganization/app/issues/1524#issuecomment-700260154

matthewblain:
The inconsistent results should be much less likley now as the entire client configuration is passed to the server, and in the typical UI flow there will be an update which should write the correct value.

Next fixes:
 * Add a transaction to the server so an error will be reported on 'collision.' 
 * Have the client retry on error somehow. Perhaps by recording a timestamp for last client side update and last server side sync, and occasionally validating that the two match.
 * Have the client report errors, e.g.  add an event to who_service like widget.analytics.logEvent(name: 'WhoServicePutClientSettings', parameters: {'result': response.statusCode});

matthewblain:
Also TODO at some point: remove server support for the old devicetoken and country methods.

brunobowden:
@matthewblain - closing this one as fixed by #1820. Please move other issues over to a separate bug at lower priority.

PR #1809, opened 2020-12-08
Make updateFirebaseToken synchronous

matthewblain:
To avoid a race condition where the Firebase Token and the user's
location are set simultaneously, leading to an inconsistent result
on the server, call UpdateFirebaseToken in a synchronous fashion.
Note that the RPC will only occur when the value actually changes.

Fixes issue #1524. In a crude fashion.

## Screenshots

<details>
</details>

- [ ] iOS Simulator
- [ ] iOS Device
- [ ] Android Simulator
- [x] Android Device
- [ ] other, please describe

I tested this by running the app in the debugger and seeing that it appears to have made the right calls.

## Checklist:

### DIFF
--- a/client/lib/main.dart
+++ b/client/lib/main.dart
@@ -255,5 +255,4 @@ class _MyAppState extends State<MyApp> with WidgetsBindingObserver {
   void updateFirebaseToken(Notifications ret) async {
     await ret.updateFirebaseToken();
   }
-
 }
--- a/client/lib/main.dart
+++ b/client/lib/main.dart
@@ -171,7 +171,7 @@ class _MyAppState extends State<MyApp> with WidgetsBindingObserver {
               update: (_, WhoService service, __) {
                 final ret = Notifications(service: service);
                 ret.configure();
-                ret.updateFirebaseToken();
+                updateFirebaseToken(ret);
                 return ret;
               },
             ),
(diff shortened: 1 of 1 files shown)


brunobowden:
@advayDev1 - would you mind reviewing this change from @matthewblain for the race condition. I'm wary about my limited understanding of Dart execution to know if putLocation and putDeviceToken are now sequential.

advayDev1:
> @advayDev1 - would you mind reviewing this change from @matthewblain for the race condition. I'm wary about my limited understanding of Dart execution to know if putLocation and putDeviceToken are now sequential.

This commit is a NOP and should not be submitted IMO.  You cannot synchronously wait on anything in a provider update, and this code does exactly what the old code did.  In the old code, the underlying async method was kicked off but not waited for, and in the new code, the wrapper method is kicked off but not waited for.

You're going to have to bite the bullet and implement transactions on the server and full-client-updates on the client.  Either solution on its own will not work in certain edge cases.  You can game out all the different orderings of memcache reads/writes, DB read/writes and network traffic from the client to prove this to yourself.

matthewblain:
OK, I'll abandon this.

brunobowden:
Thanks Advay. I had my suspicions but appreciate a definitive opinion from
you.

PR #1820, opened 2020-12-11
WIP: Client race condition fixes (putLocation and putDeviceToken in single request )

matthewblain:
This change set includes Client and Server changes to obsolete putDeviceToken and putLocation in favor of new putNotificationSettings.

It also adds a tiny bit of debug logging on the server.

The behavior should be mostly the same as before in terms of when RPCs are fired. One  RPC is skipped: if the user has not selected a country, they will not be subscribed to notifications.

No true locking is present either client or server side at this time: it is possible for up to three (or more) requests to come in 'simulteanously', where the last one will win: getstats for a new user, and a putNotificationSettings if the response is taking some period of time. However, now that the initial 'firebase token plus no country' RPC is gone, it is highly unlikley for this to happen, as it would mean that the user would be taking multiple UI actions faster than the server is responding.

A future change could include server-side transactions.

Addresses issue #1524 

## Screenshots

<details>
</details>

- [ ] iOS Simulator
- [ ] iOS Device
- [ ] Android Simulator
- [x] Android Device
- [x] other, please describe

Watching requests and checking datastore manually on my own instance of the server.

## Checklist:

### DIFF
--- a/api/who/who.proto
+++ b/api/who/who.proto
@@ -16,7 +16,7 @@ service WhoService {
   // The client should not pass a precise location.
   rpc putLocation(PutLocationRequest) returns (Void);
   
-  rpc putNotificationSettings(PutNotificationSettingsRequest) returns (Void);
+  rpc putClientSettings(PutClientSettingsRequest) returns (Void);
 
   rpc getCaseStats(GetCaseStatsRequest) returns (GetCaseStatsResponse);
 }
@@ -30,11 +30,11 @@ message PutLocationRequest {
   string isoCountryCode = 12;
 }
 
-message PutNotificationSettingsRequest {
+message PutClientSettingsRequest {
   // Firebase Cloud messaging registration token
   string token = 1;
   
-  // Approximate location. The client should not pass a precise location.
+  // User's selected location, ISO 3166-1 alpha-2 (uppercase) code.
   string isoCountryCode = 12;
 }
 
--- a/api/who/who.proto
+++ b/api/who/who.proto
@@ -15,6 +15,8 @@ service WhoService {
 
   // The client should not pass a precise location.
   rpc putLocation(PutLocationRequest) returns (Void);
+  
+  rpc putNotificationSettings(PutNotificationSettingsRequest) returns (Void);
 
   rpc getCaseStats(GetCaseStatsRequest) returns (GetCaseStatsResponse);
 }
@@ -25,6 +27,14 @@ message PutDeviceTokenRequest {
 
 message PutLocationRequest {
   // ISO 3166-1 alpha-2 (uppercase) code
+  string isoCountryCode = 12;
+}
+
+message PutNotificationSettingsRequest {
+  // Firebase Cloud messaging registration token
+  string token = 1;
+  
+  // Approximate location. The client should not pass a precise location.
   string isoCountryCode = 12;
 }
 
--- a/client/lib/api/notifications.dart
+++ b/client/lib/api/notifications.dart
@@ -105,8 +105,7 @@ class Notifications {
   Future setFirebaseToken(String newToken) async {
     var existingToken = await _userPrefs.getFirebaseToken();
     var countryCode = await prefs.countryIsoCode;
-    // If countryCode has not yet been set, no need to update the server.
-    if (existingToken != newToken && countryCode != null) {
+    if (existingToken != newToken) {
       await service.putClientSettings(
           token: newToken, isoCountryCode: countryCode);
       await _userPrefs.setFirebaseToken(newToken);
--- a/client/lib/api/notifications.dart
+++ b/client/lib/api/notifications.dart
@@ -107,7 +107,7 @@ class Notifications {
     var countryCode = await prefs.countryIsoCode;
     // If countryCode has not yet been set, no need to update the server.
     if (existingToken != newToken && countryCode != null) {
-      await service.putNotificationSettings(
+      await service.putClientSettings(
           token: newToken, isoCountryCode: countryCode);
       await _userPrefs.setFirebaseToken(newToken);
     }
--- a/client/lib/api/notifications.dart
+++ b/client/lib/api/notifications.dart
@@ -107,7 +107,8 @@ class Notifications {
     var countryCode = await prefs.countryIsoCode;
     // If countryCode has not yet been set, no need to update the server.
     if (existingToken != newToken && countryCode != null) {
-      await service.putNotificationSettings(token: newToken, isoCountryCode: countryCode);
+      await service.putNotificationSettings(
+          token: newToken, isoCountryCode: countryCode);
       await _userPrefs.setFirebaseToken(newToken);
     }
   }
--- a/client/lib/api/notifications.dart
+++ b/client/lib/api/notifications.dart
@@ -2,12 +2,14 @@ import 'package:firebase_messaging/firebase_messaging.dart';
 import 'package:flutter/widgets.dart';
 import 'package:notification_permissions/notification_permissions.dart';
 import 'package:who_app/api/user_preferences.dart';
+import 'package:who_app/api/user_preferences_store.dart';
 import 'package:who_app/api/who_service.dart';
 
 class Notifications {
   final WhoService service;
+  final UserPreferencesStore prefs;
 
-  Notifications({@required this.service});
+  Notifications({@required this.service, @required this.prefs});
 
   final FirebaseMessaging _firebaseMessaging = FirebaseMessaging.instance;
   final UserPreferences _userPrefs = UserPreferences();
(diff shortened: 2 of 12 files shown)


brunobowden:
Thanks Matthew. I'll look later today.

brunobowden:
@matthewblain - here's the command to format the code:
```
npx --no-install prettier -c '../**/*.{md,java,js,sh,yaml,yml}|*.{md,java,js,sh,yaml,yml}'
```

matthewblain:
OK; reformatted everything except the files generated by protoc. Not sure what to do that--even after reformatting there is significant change.

server/appengine/src/main/java/who/WhoServiceImpl.java
```diff
@@ -60,6 +62,28 @@ public Void putLocation(PutLocationRequest request) throws IOException {
     return new Void();
   }
 
+  @Override
+  public Void putNotificationSettings(PutNotificationSettingsRequest request)
+    throws IOException {
+    // TODO: Consider doing this in some form of datastore transaction. THe trick being that the firebase and datastore may get out of sync...
+    Client client = Client.current();
+
+    // The underlying "Topics" system allows for a null country code; would that be OK to allow here too? ?????????????????????????????????????????????
```
brunobowden, line 71: `ZZ` is a valid alpha2 code that's user assigned, so we could use that:
https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2#Decoding_table

brunobowden:
@advayDev1 - I think this is correct but I'd appreciate your perspective on it.

brunobowden:
@matthewblain - we have a policy of running the linter over the generated code. This command should fix the build failures you're seeing:
```
flutter format lib/**
```

api/who/who.proto
```diff
@@ -15,6 +15,8 @@ service WhoService {
 
   // The client should not pass a precise location.
   rpc putLocation(PutLocationRequest) returns (Void);
+  
+  rpc putNotificationSettings(PutNotificationSettingsRequest) returns (Void);
```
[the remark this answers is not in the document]
matthewblain: OK. This is also a sort of design question about intent.

api/who/who.proto
```diff
@@ -28,6 +30,14 @@ message PutLocationRequest {
   string isoCountryCode = 12;
 }
 
+message PutNotificationSettingsRequest {
+  // Firebase Cloud messaging registration token
+  string token = 1;
+  
+  // Approximate location. The client should not pass a precise location.
```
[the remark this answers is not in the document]
matthewblain: Yes. I copy/pasted this comment from above. Design question: is this the correct granulatiry? California is different from Florida, but both are US.

server/appengine/src/main/java/who/WhoServiceImpl.java
matthewblain: The system uses null/"" as "country not set" elsewhere on client and server, which seems right to me. The question here is a design question: what is the intent of this API.

matthewblain:
Updated to reflect suggestions. Retested against my deploy by editing endpoints.dart.
You'll note there are some design questions here. I have written out my assumptions here:
https://docs.google.com/document/d/1Itu46Qz43EVw2vasY0YvYDWydAlUyqMAEUgtRcROTGA/edit

client/lib/api/notifications.dart
```diff
@@ -97,16 +99,16 @@ class Notifications {
...
   }
 
   Future setFirebaseToken(String newToken) async {
-    var exitsingToken = await _userPrefs.getFirebaseToken();
-    if (exitsingToken != newToken) {
-      await service.putDeviceToken(newToken);
+    var existingToken = await _userPrefs.getFirebaseToken();
+    var countryCode = await prefs.countryIsoCode;
+    // If countryCode has not yet been set, no need to update the server.
+    if (existingToken != newToken && countryCode != null) {
```
brunobowden, line 109: Shouldn't this update the token even if the countryCode is null? If the `countryCode` is null mistakenly, I'd rather see that as an error server side rather than hidden on the client.

If anything is silently failing, better to add a `analytics.logEvent` to capture it.

client/lib/api/who_service.dart
```diff
@@ -12,32 +12,18 @@ class WhoService {
 
   WhoService({@required String endpoint}) : serviceUrl = '$endpoint/WhoService';
 
-  /// Put device token.
-  Future<bool> putDeviceToken(String token) async {
+  /// Put Notification Settings
```
brunobowden, line 15:
```suggestion
  /// Put Client Settings
```

client/lib/pages/onboarding/onboarding_page.dart
```diff
@@ -118,11 +118,12 @@ class _OnboardingPageState extends State<OnboardingPage> {
...
     await _toNextPage();
+    var fcmToken = await UserPreferences().getFirebaseToken();
     try {
-      await widget.service
-          .putLocation(isoCountryCode: _selectedCountry.alpha2Code);
+      await widget.service.putClientSettings(
+          token: fcmToken, isoCountryCode: _selectedCountry.alpha2Code);
     } catch (error) {
-      print('Error sending location to API: $error');
+      print('Error sending client settings to API: $error');
```
brunobowden, line 126: Ideally this should log an error in Analytics but lower priority for now.

```
widget.analytics.logEvent(....)
```

server/appengine/src/main/java/who/WhoServiceImpl.java
```diff
@@ -60,6 +62,28 @@ public Void putLocation(PutLocationRequest request) throws IOException {
...
 
+  @Override
+  public Void putClientSettings(PutClientSettingsRequest request)
+    throws IOException {
+    // TODO: Consider doing this in some form of datastore transaction. THe trick being that the firebase and datastore may get out of sync...
+    Client client = Client.current();
+
+    // The underlying "Topics" system allows for a null country code; would that be OK to allow here too? ?????????????????????????????????????????????
+    if (
+      request.isoCountryCode == null ||
```
brunobowden, line 73: Update null check if needed

brunobowden (approved):

server/appengine/src/main/java/who/WhoServiceImpl.java
```diff
@@ -60,6 +62,28 @@ public Void putLocation(PutLocationRequest request) throws IOException {
...
+      request.isoCountryCode.length() != 2 ||
```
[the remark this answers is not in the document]
matthewblain: Yes. It was there before and seems reasonable. Though I don't know the behavior of java regexes etc--presumably it should terminate at 2 no matter what?

api/who/who.proto
```diff
@@ -28,6 +30,14 @@ message PutLocationRequest {
   string isoCountryCode = 12;
 }
 
+message PutClientSettingsRequest {
+  // Firebase Cloud messaging registration token
+  string token = 1;
+  
+  // User's selected location, ISO 3166-1 alpha-2 (uppercase) code.
+  string isoCountryCode = 12;
```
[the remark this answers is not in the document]
matthewblain: The user may or may not have selected a countrycode; seems simplest to just allow it to be "". Quite likley it can never actually be 'null', I haven't checked the docs.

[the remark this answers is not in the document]
brunobowden: proto3 is reintroducing optional fields but they end up being quite a hassle. Easier to allow "" and then be done with it.

That means fixing the client code to report an empty string when the country isn't set..... or just set country to be a default of empty string.

client/lib/api/notifications.dart
matthewblain: Yeah this comment may be wrong. I think it's probably better to have it not trigger until onboarding is complete?
I'm trying to avoid having two requests in flight if the token and countrycode 'show up' at the same time. That's one of the two possible race conditions in the current system, since these two events are async. That's why I have lots of open questions in the doc I shared!

client/lib/pages/onboarding/onboarding_page.dart
matthewblain: Good idea. Maybe in who_service?
 widget.analytics.logEvent(
                        name: 'WhoServicePutClientSettings',
                        parameters: {'result': response.statusCode});

More important however is having the client know it should retry at some point. When should that be done? A flag could be saved in the prefs need_server_update, then at some time (e.g. the next launch it would know that it needs to try again.). Heck, that could reduce the 'race condition' by doing that as some sort of delayed event on the client. (E.g. save prefs; trigger async server update) But that seems like it should be a future change.

matthewblain:
I removed the null check. Indeed if you pass a null to the proto marshaller it will raise an error. 
However, this should never happen, as the ordering is Onboarding->Select Country->Enable Notifications. And there's no way in the UI to undo this. Perhaps it's possible to edit the underlying user prefs database somehow, but I don't know that the app is designed to handle edits of that sort beyond crashing.

brunobowden:
Bottom line is that the app shouldn't ever crash but instead fail gracefully and record an error.... but now I'm second guessing myself. Once you're comfortable, please go ahead and land this.

PR #1829, opened 2020-12-15
Analytics Account Settings data sharing

brunobowden:
Make repo consistent with existing settings
on production analytics account.

## Checklist:

### DIFF
--- a/server/terraform/README.md
+++ b/server/terraform/README.md
@@ -232,7 +232,7 @@ Analytics may be disabled by the user on the client app on the Settings page.
    1. Configure "Country of Business" (doesn't change where data is processed)
       - Dev: United States
       - Prod: Switzerland
-   1. For "Data Sharing Settings", disable "Google products & services"
+   1. For "Data Sharing Settings", enable "Google products & services"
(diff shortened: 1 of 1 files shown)


yukuairoy (approved):

advayDev1:
This is my personal opinion and does not reflect the project or WHO.

Nothing herein is legal advice: The privacy policy should reflect what sharing is going on of App's user data and Google for the purposes of its own products and service.  Furthermore, Google requires agreement to various terms when using this check box. 

Screen Shot 2020-12-15 at 1 12 17 PM


An initial design goal in using Google Analytics was to keep analytics data as private as possible.   It also affects things like demographic and interest collection which again the initial design goal was to never collect such information.

advayDev1:
Screenshot 2020-12-17 at 7 49 24 PM
And is this description accurate if GA gets used for more than what it says?

brunobowden:
Closing this as it will not be changing. WHO has confirmed that they've disabled sharing with "Google products & services". So the repo documentation now matches production on this setting. Still working on verifying other settings.

PR #1840, opened 2020-12-18
Stats update caching

### LINKED ISSUE #1835: App makes too many queries to the backend
## Summary

When the app is running, it makes many queries to GetCaseStats, and several for the static content bundles.

These values should be cached locally, possibly with an expiration from the server, but certainly for many hours absent an explicit user Refresh request.

## Reproduction Steps

Open the app. 
Browse around on it, switch to another app and back, etc. Don't change settings or otherwise take action to explicitly interact with the server.

## Expected Behavior

One, or perhaps zero, request to the backend for each type of server provided data.

## System

- Device: Nexus 7
- OS: Android
- App Version: [1.2.0+14]

## Additional Context

When I tested this, in 5 minutes I got 13 requests to GetCaseStats, and one or two to each of 16 different static files.

I don't know if this could be Android-version-specific, but I expect that it is universal.

See attached csv file with requests seen by my dev instance over my 5 minute session.

[getcasestatslogs.txt](https://github.com/WorldHealthOrganization/app/files/5701402/getcasestatslogs.txt)

## Checklist:

- [ ] Searched the [existing issues](https://github.com/WorldHealthOrganization/app/issues) to ensure you are not creating a duplicate.
- [ ] Followed the [Contributor Guidelines](https://github.com/WorldHealthOrganization/app/blob/master/docs/CONTRIBUTING.md).

matthewblain:
The StatsStore now tracks the country and time that the last stats loaded .
It now will not (re)load stats if the country is unset, or if the country is the same but the time has been too long.

Ideally, this system would be replaced with a single load and something like the RefreshIndicator class, but that is a much larger change.

Fixes #1835

- [ ] iOS Simulator
- [ ] iOS Device
- [ ] Android Simulator
- [x] Android Device
- [ ] other, please describe

Ran app and monitored the debug logs for lines beginning with `StatsStore update`. Verified that the logs indicate that the initial GetCaseStats is skipped, that the first one with a country is loaded, that subsequent ones are not reloaded if it is too soon.
Changed country in UI and noted that the stats are reloaded. Waited for timeout to pass and noted that the stats are reloaded.

## Checklist:


brunobowden (approved):
Looking good to me

PR #1844, opened 2020-12-19
Re-TOS and re-Onboard all users

### LINKED ISSUE #1842: Re-ToS/PN users with the final version of the launch docs
@brunobowden Hunter made a change a while back re-tosing users, which I guess can be redone again for the launch version.  Most importantly GA opt-outs should be preserved across re-toses which is a bit tricky.  Think we discussed this a few months ago, but I don't see a bug.

Nothing herein is legal advice.

brunobowden:
- OnboardingCompletedV2 repeats same process as in #1450
- Increasingly unwieldy but the safest approach since used previously
- `LegalCompletedTimestamp` recorded for facilitating new approach
- Fixes #1842

- [x] Android Simulator

Repeated onboarding and fresh onboarding

## Checklist:

### DIFF
--- a/client/lib/api/user_preferences.dart
+++ b/client/lib/api/user_preferences.dart
@@ -3,7 +3,6 @@ import 'dart:math';
 import 'package:firebase_analytics/firebase_analytics.dart';
 import 'package:firebase_messaging/firebase_messaging.dart';
 import 'package:shared_preferences/shared_preferences.dart';
-
 import 'package:uuid/uuid.dart';
 import 'package:uuid/uuid_util.dart';
 import 'package:who_app/main.dart';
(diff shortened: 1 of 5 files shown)


client/lib/api/user_preferences.dart
```diff
@@ -59,15 +58,39 @@ class UserPreferences {
...
+  Future<bool> getLegalCompleted() async {
     return (await SharedPreferences.getInstance())
-            .getBool(UserPreferenceKey.TermsOfServiceCompleted.toString()) ??
+            .getBool(UserPreferenceKey.LegalCompleted.toString()) ??
         false;
   }
 
-  Future<bool> setTermsOfServiceCompleted(bool value) async {
+  Future<bool> setLegalAccepted() async {
+    var now = DateTime.now().millisecondsSinceEpoch;
```
advayDev1, line 81: Tbh its unclear what use a local time is for definitive purposes.  You'd have to get a trusted timestamp say from a server for it to be meaningful.

advayDev1:
Nothing herein is legal advice.  I assume you will wait until the final TOS/PN are published by WHO before committing this?

client/lib/api/user_preferences.dart
advayDev1: More specifically, it seems by a new approach you want to use this to determine whether to re-tos ppl in the future and a local time cannot tell you which tos/pn was accepted.  You could for example, associate all re-ToS/PNs with a new build version and store the app's build number at time of acceptance.  Or you could pull down the ToS/PN via HTTPS and store their combined hash or something.  Not legal advice.

brunobowden: I see your point. This whole system needs to be reworked and doing it right is beyond what we can do in time for the launch.

theswerd (approved):
LGTM 

If we are switching **getOnboardingCompletedV1** to **getOnboardingCompletedV2** wouldn't it be better to delete **getOnboardingCompletedV1** and replace it with ** getOnboardingCompleted**?

client/lib/pages/onboarding/onboarding_page.dart
```diff
@@ -128,10 +128,11 @@ class _OnboardingPageState extends State<OnboardingPage> {
   }
 
   Future<void> _onLegalDone(ContentStore store) async {
-    await UserPreferences().setTermsOfServiceCompleted(true);
+    await UserPreferences().setLegalAccepted();
     // Enable auto init so that analytics will work
     await FirebaseMessaging.instance.setAutoInitEnabled(true);
     if (!await UserPreferences().getOnboardingCompleted() ||
+        !await UserPreferences().getOnboardingCompletedV1() ||
```
advayDev1, line 135: hmmm this logic may not be right when you add this.
consider when onboardingComplete=false, and onboardingCompetedV1=true, and analyticsEnabled=false
this will still analytics on, which is wrong

brunobowden: Good catch. Fixed this logic:

1. Only enable analytics for first onboarding.... i.e. when both prior onboardings are false
1. Dropped `|| await UserPreferences().getAnalyticsEnabled()` in logic as enabling analytics when it's already enabled is completely redundant

brunobowden:
`getOnboardingCompletedV1` is being left as is.... but the diff makes it look likes it's being replaced. Ultimately this whole system needs to be ripped out and replaced with something better.

PR #1863, opened 2020-12-20
GetStats: Persist the country only on success

### LINKED ISSUE #1858: Wrong Stats are being displayed when internet connectivity is lost and country is changed
1.  Launch the app and complete onboarding journey.
2.  Tap on Stats icon in the footer.
3.  Stats for the selected country is displayed to the user.
4.  Switch of mobile internet.
5.  Go to Settings and and change the country from the dropdown.
6.  Tap on Stats again after changing the country.

Expected Behavior
Previous Country Stats should not be loaded for the changed country

Actual Behavior 
Stats of previously selected country are being displayed for the newly selected country.

 System

- Device: Huawei Honor 7c
- OS: Android 8.0
- App Version: Version 1.0.0-staging(281)

 Additional Context

Another observation is that if we change the country by the search option in the dropdown then this bug doesn't occur, it only happens when we change the country by scrolling the country selector dropdown.

https://user-images.githubusercontent.com/63721227/102717147-706d2d00-4306-11eb-9d45-2a5755048e5d.mp4

Checklist:

matthewblain:
GetStats: Persist the country only on success.
Fixes #1858 

## Screenshots

<details>
</details>

- [ ] iOS Simulator
- [ ] iOS Device
- [ ] Android Simulator
- [ ] Android Device
- [ ] other, please describe

I have not yet tested this! Will do so later today.

## Checklist:

### DIFF
--- a/client/lib/api/stats_store.dart
+++ b/client/lib/api/stats_store.dart
@@ -74,21 +74,21 @@ abstract class _StatsStore with Store implements Updateable {
 
     var now = DateTime.now().millisecondsSinceEpoch;
     var delta = now - _lastUpdateTimestamp;
-    if (delta < statsUpdateFrequency &&
-        countryIsoCode == prefs.countryIsoCode) {
+    var countryToLoad = prefs.countryIsoCode;
+
+    if (delta < statsUpdateFrequency && countryIsoCode == countryToLoad) {
       print('StatsStore update skipped, cached data used from ${delta}ms ago.');
       return;
     }
 
-    var countryToLoad = prefs.countryIsoCode;
     if (countryToLoad == null) {
       // This happens at startup time (while the prefs are being loaded in?).
       print('StatsStore update skipped, country unknown.');
       return;
     }
 
-    print('StatsStore update with country ${countryIsoCode}, '
-        'refreshing data last seen ${delta}ms ago.');
+    print('StatsStore update with country ${countryToLoad}, '
+        'refreshing data last seen ${delta}ms ago for ${countryIsoCode}.');
 
     caseStatsResponse =
         await service.getCaseStats(isoCountryCode: countryToLoad);
@@ -96,6 +96,5 @@ abstract class _StatsStore with Store implements Updateable {
     // persisted values will not be updated.
     _lastUpdateTimestamp = now;
     countryIsoCode = countryToLoad;
-
   }
 }
--- a/client/lib/api/stats_store.dart
+++ b/client/lib/api/stats_store.dart
@@ -80,8 +80,8 @@ abstract class _StatsStore with Store implements Updateable {
       return;
     }
 
-    countryIsoCode = prefs.countryIsoCode;
-    if (countryIsoCode == null) {
+    var countryToLoad = prefs.countryIsoCode;
+    if (countryToLoad == null) {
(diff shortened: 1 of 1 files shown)


client/lib/api/stats_store.dart
```diff
@@ -80,8 +80,8 @@ abstract class _StatsStore with Store implements Updateable {
       return;
     }
 
-    countryIsoCode = prefs.countryIsoCode;
-    if (countryIsoCode == null) {
+    var countryToLoad = prefs.countryIsoCode;
+    if (countryToLoad == null) {
```
brunobowden, line 84: How about inline `prefs.countryIsoCode`?
```suggestion
    if (prefs.countryIsoCode == null) {
        caseStatsResponse =
            await service.getCaseStats(isoCountryCode: prefs.countryIsoCode);	        
        _lastUpdateTimestamp = now;
        countryIsoCode = prefs.countryIsoCode;
    }
```

brunobowden:
git precommit checks aren't running so try:
```
flutter format lib/**
```

client/lib/api/stats_store.dart
brunobowden: On second thoughts, the local variable avoids a possible race condition where the country setting was changed

brunobowden (approved):

matthewblain:
Fixed formatting. Small debuglog and other tweaks.

Tested: Disabled internet / switched country / enabled as per bug. 
The 'stale' data from the old country will remain until the internet connection returns, then will quickly refresh to the selected country.
It would be possible to zero out the data on country change, but I don't know if it's better to show blank data there or 'wrong' data. It is labeled correctly (with the 'old' country) in the UI.

Triage/priority note: without this fix, the problem will resolve itself when the local data expires, either on app restart or 1 hour.

yukuairoy (approved):

Mohit020888:
> Fixed formatting. Small debuglog and other tweaks.
> 
> Tested: Disabled internet / switched country / enabled as per bug.
> The 'stale' data from the old country will remain until the internet connection returns, then will quickly refresh to the selected country.
> It would be possible to zero out the data on country change, but I don't know if it's better to show blank data there or 'wrong' data. It is labeled correctly (with the 'old' country) in the UI.
> 
> Triage/priority note: without this fix, the problem will resolve itself when the local data expires, either on app restart or 1 hour.

@matthewblain - I think it's better to show blank data suggesting Internet connectivity issues to the user rather than populating wrong data.

brunobowden:
@matthewblain @Mohit020888 - this is somewhat a corner case. I'd rather invest energy in to migrating to Firestore which will handle this situation and many others far better.

theswerd (approved):
LGTM - should we add an indicator that the network failed?

brunobowden:
@theswerd - I think people generally know when they don’t have an internet connection. I’d rather they invest the effort in migrating to Firestore

matthewblain:
OK I'll merge this now. Agree that this is a corner case, even this fix itself, and firebase is likley to provide a more robust mechanism.

Mohit020888:
@matthewblain @brunobowden - being a tester I can't help myself playing with these corner case scenarios and reporting them but I agree with prioritizing other important tasks which can take care of these issues.

PR #1864, opened 2020-12-20
Respect Disable Notifications User Setting

### LINKED ISSUE #1857: Notifications disable doesn't work (topics issue?)
## Summary

A clear and concise description of what the problem is, from the user's perspective

## Reproduction Steps

1. Go to Settings
3. Enable "Notifications"
4. Send message to `ALL~LOCATION~US` topic
5. WORKS: notification received

6. Disable "Notifications"
4. Send message to `ALL~LOCATION~US` topic
5. FAILS: notification received

## Expected Behavior

No notifications

## System

- Nexus 5
- Android 6.0.1

## Checklist:

brunobowden:
- Notification disabled implemented by setting token to zero
- `updateSubscriptions` on BE will unsubscribe all topics in that case
- Fixes #1857
- TODO: more testing

- Staging server and Android device
- Toggling on and off notifications
- Sending to both `ALL~LOCATION` and `ALL~LOCATION~US` topics

## Checklist:

### DIFF
--- a/server/appengine/src/main/java/who/Client.java
+++ b/server/appengine/src/main/java/who/Client.java
@@ -22,6 +22,7 @@ public class Client {
 
   /** Device token for notifications. */
   public String token;
+  public boolean disableNotifications;
 
   public Platform platform;
 
--- a/server/appengine/src/main/java/who/NotificationsManager.java
+++ b/server/appengine/src/main/java/who/NotificationsManager.java
@@ -35,9 +35,7 @@ public class NotificationsManager {
 
   public void updateSubscriptions(Client client) throws IOException {
     Preconditions.checkNotNull(client);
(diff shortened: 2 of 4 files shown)


yukuairoy (approved):

matthewblain (approved):
Seems OK. This will persist the old token in the datastore, but that should be harmless. I don't understand what happens (on the firebase side) if the token changes from XYZ123 to ABC987, but presumably it should transition through null first?

brunobowden:
Disabling and re-enabling notifications means the Firebase client library will share the same token again. I believe the token might become invalid for other reasons though and get changed that way. If it does change, then overwriting it should be fine as I'd expect the old token to be invalid.

brunobowden:
Pushed to all servers

PR #1878, opened 2020-12-24
QA Buttons: crashing and notifications

### LINKED ISSUE #1779: QA Buttons: crashing and notifications
## Summary

For QA and Debugging it would be helpful to have several manual buttons available in the non production build:

1. Crash button to generate crash log: `FirebaseCrashlytics.instance.crash();`
2. Send notification to user. Could require server API that's disable in production to prevent abuse.
3. Anything else?

## Checklist:

theswerd:
Closes #1779 

- [x] iOS Simulator
- [ ] iOS Device
- [ ] Android Simulator
- [ ] Android Device
- [ ] other, please describe

## Checklist:

### DIFF
--- a/server/appengine/src/main/java/who/WhoServiceImpl.java
+++ b/server/appengine/src/main/java/who/WhoServiceImpl.java
@@ -13,6 +13,9 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import present.rpc.ClientException;
 
+import com.google.firebase.messaging.FirebaseMessaging;
+
+
 public class WhoServiceImpl implements WhoService {
 
   NotificationsManager nm;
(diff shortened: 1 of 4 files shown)


server/appengine/src/main/java/who/WhoServiceImpl.java
```diff
@@ -139,4 +142,18 @@ public GetCaseStatsResponse getCaseStats(GetCaseStatsRequest request)
       .ttl(STATS_TTL_SECONDS)
       .build();
   }
+
+  @Override
+  public Void sendTestNotification(SendTestNotificationRequest request) throws IOException {    
+    Message message = Message.builder()
```
matthewblain, line 148: What is the security model here? 
One possible way to enable this sort of test button :
Use the client's token (from the datastore). 

Check if the client has a bit set to enable this feature, and only send if the bit is set.
(Now the question becomes how to set the bit, which depends on the goal here. )

brunobowden: I think the best is to get the client to supply the token in the request. The point is that if the requester has the token... then they've got he secret anyway. The server side actions should be:

1. If request token doesn't match user token in the database, then report error (should be displayed to use in client). No server writes to Datastore.
2. Send message to only that user and no other. So you can only spam yourself and not anyone else.

I think that makes it secure enough. Not sure if we should also add rate limiting on the API. It should only get a low number of QA interactions and not a large volume. I think that would be simpler than a per user rate limit.

brunobowden: Goal is to allow for easier QA and avoid it being possible to spam anyone else. Being able to spam yourself is ok... unless that can be used to exhaust quota or trigger some kind of other issue.

PR #1879, opened 2020-12-24
Added alert dialog when attempting to close

### LINKED ISSUE #1228: Pressing back button from any footer functionalities exits the app without prompting user.
Is your feature request related to a problem? Please describe.
When user finishes the onboarding permissions and lands on application homescreen, then pressing the back button from Stats, Learn, Check-up , Settings directly kills the app without giving a prompt to the user.

Describe the solution you'd like
If the user is on any other functionality other then Home functionality in the footer then pressing the back button from mobile should take the user back to the Home screen tab and further press of back button should give a pop up to the user asking "Are you sure you want to exit the app?" with okay and cancel button.

- [ x] Searched the [existing issues](https://github.com/WorldHealthOrganization/app/issues) to ensure you are not creating a duplicate.
- [ x] Followed the [Contributor Guidelines](https://github.com/WorldHealthOrganization/app/blob/master/docs/CONTRIBUTING.md).
- [ x] Issue has descriptive title focused on the feature or problem from the user's perspective

theswerd:
Closes #1228

## Checklist:

### DIFF
--- a/client/lib/pages/main_pages/home_page.dart
+++ b/client/lib/pages/main_pages/home_page.dart
@@ -56,14 +56,50 @@ class HomePage extends ContentWidget<IndexContent> {
       ];
     }
 
-    return PageScaffold(
-      showHeader: false,
-      // For background scroll bleed only - white background set on _HomePageSection widgets
-      color: content?.items != null
-          ? Constants.primaryDarkColor
-          : CupertinoColors.white,
-      beforeHeader: _buildPromo(),
-      body: _buildBody(context),
+    return WillPopScope(
+      child: PageScaffold(
+        showHeader: false,
+        // For background scroll bleed only - white background set on _HomePageSection widgets
+        color: content?.items != null
+            ? Constants.primaryDarkColor
+            : CupertinoColors.white,
+        beforeHeader: _buildPromo(),
+        body: _buildBody(context),
+      ),
+      onWillPop: () async {
+        bool willPop;
+        
+        await showCupertinoDialog(
+          context: context,
+          builder: (context) => CupertinoAlertDialog(
+            title: Text('Trying to exit?'),
+            content: Text('Are you sure you want to exit the app?'),
+            actions: [
+              CupertinoDialogAction(
+                child: Text(
+                  'Exit',
+                ),
+                onPressed: () {
+                  willPop = true;
+                  Navigator.pop(context);
+                },
+              ),
+              CupertinoDialogAction(
+                child: Text(
+                  'Cancel',
+                ),
+                onPressed: () {
+                  willPop = false;
+                  Navigator.pop(context);
+                },
+                isDefaultAction: true,
+              ),
+            ],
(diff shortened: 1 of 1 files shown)


theswerd:
@brunobowden @yukuairoy Can you guys take a look at this?

yukuairoy (approved):

brunobowden:
@theswerd @Mohit020888 - sorry to be stepping later on this. How do we know what the correct behavior should be? In all my years, I've never seen an app give me the message "Do you want to exit this app?".... so it's not clear to me that this is the right answer. The best thing would be a commonly used app that has behavior like this.... but that might be harder to find as you have to test on an old version of Android? It's just that I'm not familiar enough with the Android UI to evaluate this.

@theswerd - please add your further thoughts on this.

theswerd:
@brunobowden I'm not an Android guy - just built what the issue asked for. @jasonTelanoff might have some other insight into this.

jasonTelanoff:
@brunobowden there are several apps on android that would have an alert dialogue pop up when trying to exit the app by clicking the back button. This is to prevent the user from exiting the app while spamming back. Though this is more prevalent on older apps and games, I think it would be a useful feature. The only downside I've encountered and can think of are that user trying to exit the app through the back button need to tap again. This is easily avoidable however by clicking the home button or exiting some other way. I believe this should be included in the app, as the benefits outweigh the costs, but it is not a top priority or absolutely necessary; it's just a nice feature to include for us Android users.

Mohit020888:
> @theswerd @Mohit020888 - sorry to be stepping later on this. How do we know what the correct behavior should be? In all my years, I've never seen an app give me the message "Do you want to exit this app?".... so it's not clear to me that this is the right answer. The best thing would be a commonly used app that has behavior like this.... but that might be harder to find as you have to test on an old version of Android? It's just that I'm not familiar enough with the Android UI to evaluate this.
> 
> @theswerd - please add your further thoughts on this.

@brunobowden -The app should atleast take the user to the home functionality in the footer when the user is on any other footer functionality like stats, learn , check-up,settings etc. and presses mobile back button.
Further adding a message that another press of back button will make the user exit the app is also very useful feature and you can see this behavior in the following example of my mobile carrier app here in India 

https://user-images.githubusercontent.com/63721227/103436463-8dc4c400-4c41-11eb-998f-36ed9307fa49.mp4

brunobowden:
@theswerd - see the video that @Mohit020888 showed. That looks like a good way of handling the experience. Instead of a dialog where someone has to confirm they want an exit (which is irritating in having to move your finger to press it).... have a popup as shown in the Airtel app and then pressing the back button twice exits the app. I think mimicking the same behavior would be the best approach.

brunobowden:
@theswerd - if the functionality matches what you see in [@Mohit020888's video](https://github.com/WorldHealthOrganization/app/pull/1879#issuecomment-753292254), then I'm good for you to merge this.

PR #1880, opened 2020-12-26
Make RefreshCaseStats testable

matthewblain:
Make RefreshCaseStats testable and add a skeletal test.
* Refactor RefreshCastStatsServlet.
* Add RefreshCaseStatsServletTest
* Switch from deprecated JsonParse.parse to JsonParse.parseString.
* Removed n>100000 limit. (Not sure why it was added.)
* Make WhoTestSupport work w/that test.
* Fix bug in Environment with app engine appid shard.

## Screenshots

 n/a

- [ ] iOS Simulator
- [ ] iOS Device
- [ ] Android Simulator
- [ ] Android Device
- [x] other, please describe

Ran the new unit test and it passed. Verfiied that making the expected value something other than 2 causes it to fail.

## Checklist:

### DIFF
--- a/server/appengine/src/main/java/who/Environment.java
+++ b/server/appengine/src/main/java/who/Environment.java
@@ -51,10 +51,13 @@ public enum Environment {
 
   private static String getApplicationId() {
     String applicationId = AppEngine.applicationId();
-    // "o~" prefix bug in App Engine. Not sure where those 2 chars at the beginning come from.
-    final String PREFIX_BUG = "o~";
-    if (applicationId.startsWith(PREFIX_BUG)) {
-      applicationId = applicationId.substring(PREFIX_BUG.length());
+    if (applicationId == null) {
+      return null;
+    }
+    // Strip shard.
+    int tilde = applicationId.indexOf('~');
+    if (tilde >= 0) {
+      applicationId = applicationId.substring(tilde + 1);
     }
     return applicationId;
   }
--- a/server/appengine/src/main/java/who/RefreshCaseStatsServlet.java
+++ b/server/appengine/src/main/java/who/RefreshCaseStatsServlet.java
@@ -70,9 +70,7 @@ public class RefreshCaseStatsServlet extends HttpServlet {
       snapshot.totalCases = 0L;
       snapshot.totalDeaths = 0L;
       data.snapshots.put(timestamp, snapshot);
-    } else {
-      // There shouldn't be a snapshot already except for GlobalStats!!!
-    }
+    } 
 
     snapshot.dailyCases += dailyCases;
     snapshot.dailyDeaths += dailyDeaths;
--- a/server/appengine/src/main/java/who/RefreshCaseStatsServlet.java
+++ b/server/appengine/src/main/java/who/RefreshCaseStatsServlet.java
@@ -1,5 +1,6 @@
 package who;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.gson.JsonArray;
 import com.google.gson.JsonElement;
 import com.google.gson.JsonNull;
@@ -16,12 +17,14 @@ import javax.servlet.http.HttpServletResponse;
 import okhttp3.OkHttpClient;
 import okhttp3.Request;
 import okhttp3.Response;
+import org.jetbrains.annotations.NotNull;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 public class RefreshCaseStatsServlet extends HttpServlet {
 
-  private static final class JurisdictionData {
+  @VisibleForTesting
+  static final class JurisdictionData {
 
     long totalCases = 0L, totalDeaths = 0L;
     long lastUpdated = 0L;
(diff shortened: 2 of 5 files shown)


server/appengine/src/main/java/who/RefreshCaseStatsServlet.java
```diff
@@ -67,18 +70,106 @@ private static void updateData(
       snapshot.totalCases = 0L;
       snapshot.totalDeaths = 0L;
       data.snapshots.put(timestamp, snapshot);
+    } else {
+      // There shouldn't be a snapshot already except for GlobalStats!!!
```
brunobowden: Pass in global as a boolean and then produce error if this codepath is encountered for non global stats? It would mean that we have a repeated timestamp which is likely a serious problem we should now about. Fine to defer to another PR.

server/appengine/src/main/java/who/Environment.java
```diff
@@ -51,10 +51,13 @@ public static Environment current() {
...
-      applicationId = applicationId.substring(PREFIX_BUG.length());
+    if (applicationId == null) {
+      return null;
+    }
+    // Strip shard.
+    int tilde = applicationId.indexOf('~');
+    if (tilde >= 0) {
+      applicationId = applicationId.substring(tilde + 1);
     }
     return applicationId;
```
brunobowden, line 62: Extract to method, so it can be tested without needing injection.
```
assertEquals(stripShard(null), null);
assertEquals(stripShard("o~AppEngineID"), "AppEngineID");
assertEquals(stripShard("dev~AppEngineID"), "AppEngineID");
```

server/appengine/src/main/java/who/RefreshCaseStatsServlet.java
```diff
@@ -67,18 +70,104 @@ private static void updateData(
...
+    Map<String, JurisdictionData> countryData,
+    JurisdictionData globalData
+  ) {
+    // Given that each row has heterogeneous elements, not sure there is much benefit
+    // to using gson with reflection here.
+    for (JsonElement feature : rows) {
+      JsonObject featureAsJsonObject = feature.getAsJsonObject();
+      JsonElement attributesElement = featureAsJsonObject.get("attributes");
+
+      JsonObject attributes = attributesElement.getAsJsonObject();
```
brunobowden, line 100: Replace with single line?
```
JsonObject attributes = feature.getAsJsonObject().get("attributes").getAsJsonObject();
```

server/appengine/src/main/java/who/RefreshCaseStatsServlet.java
```diff
@@ -100,105 +189,30 @@ protected void doGet(
...
-          .collect(Collectors.toList())
-      )
-      .attribution("WHO");
-    StoredCaseStats.save(global.build());
+    CaseStats globalStats = buildCaseStats(
+      JurisdictionType.GLOBAL,
+      "",
+      globalData
+    );
+    StoredCaseStats.save(globalStats);
```
brunobowden, line 207: Nice to see the duplication go away....

server/appengine/src/main/java/who/RefreshCaseStatsServlet.java
```diff
@@ -100,105 +189,30 @@ protected void doGet(
...
-            country,
-            timestamp,
-            dailyDeaths,
-            totalDeaths,
-            dailyCases,
-            totalCases
-          );
-        }
-      }
+      processWhoStats(rows, countryData, globalData);
```
brunobowden, line 199: I think `processWhoStats` should include the JSON parsing as well. So it takes a single string and then does all the work. Better encapsulation for testing as well.

server/appengine/src/test/java/who/RefreshCaseStatsServletTest.java
```diff
@@ -0,0 +1,41 @@
...
+        .getResourceAsStream("smaller-who-data.json");
+    JsonObject root = JsonParser
+      .parseReader(new InputStreamReader(in, "UTF-8"))
+      .getAsJsonObject();
+
+    JsonArray rows = root.getAsJsonArray("features");
+
+    RefreshCaseStatsServlet.JurisdictionData globalData = new RefreshCaseStatsServlet.JurisdictionData();
+    Map<String, RefreshCaseStatsServlet.JurisdictionData> countryData = new HashMap<>();
+    servlet.processWhoStats(rows, countryData, globalData);
```
brunobowden, line 37: `processWhoStats` should be doing the json parsing.... instead accepting just a string for the file contents.

brunobowden (approved):

server/appengine/src/main/java/who/RefreshCaseStatsServlet.java
matthewblain: this was part of an earlier experiment.

server/appengine/src/main/java/who/Environment.java
matthewblain: I don't want to change this part of the code beyond the minimum right now. This is really a bug here:
https://github.com/presentco/present-engine/blob/master/src/main/java/present/engine/AppEngine.java#L36

server/appengine/src/main/java/who/RefreshCaseStatsServlet.java
matthewblain: I think it's the opposite. processWhoStats takes in pretty structured things as input and output. the json parsing is done beforehand. note that this allows it to be done different ways (e.g. the unit test does not actually use a string).  and the outer loop right here actually uses some of the data in the parsed json which isn't really relevant to processWhoStats.

matthewblain:
updated and added more complete (but likley to shortly be deleted) data assertion validation

brunobowden:
LGTM

PR #1889, opened 2020-12-27
Case Stats update rejected for unexpected change

brunobowden:
Case Stats update is rejected if:
- Global cases drop below prior value
- Global stats rise more than 5% + 30,000 since last update
- Skip checks for first case stats entry

Unit tests

## Checklist:

### DIFF
--- a/server/appengine/src/main/java/who/RefreshCaseStatsServlet.java
+++ b/server/appengine/src/main/java/who/RefreshCaseStatsServlet.java
@@ -168,7 +168,6 @@ public class RefreshCaseStatsServlet extends HttpServlet {
         );
       }
     }
-    fixPartialLastDayAll(countryData, globalData);
   }
 
   /**
@@ -323,7 +322,9 @@ public class RefreshCaseStatsServlet extends HttpServlet {
       numItems += rows.size();
       processWhoStats(rows, countryData, globalData);
     }
+    fixPartialLastDayAll(countryData, globalData);
 
+    // Reject unexpected changes in case stats, e.g. too large an increase
     CaseStats oldCaseStats = StoredCaseStats.load(JurisdictionType.GLOBAL, "");
     if (oldCaseStats != null) {
       totalCasesDeltaCheck(oldCaseStats.cases, globalData.totalCases);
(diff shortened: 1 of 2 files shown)


yukuairoy (approved):

server/appengine/src/test/java/who/RefreshCaseStatsServletTest.java
```diff
@@ -219,4 +219,17 @@ public void testDec24Data() throws IOException {
     RefreshCaseStatsServlet.JurisdictionData ng = countryData.get("RW");
     assertEquals(356, ng.snapshots.size());
   }
+
+  @Test
+  public void totalCasesDeltaCheck() {
+    RefreshCaseStatsServlet servlet = new RefreshCaseStatsServlet();
+    servlet.totalCasesDeltaCheck(1000, 2000);
+    servlet.totalCasesDeltaCheck(77_000_000, 78_000_000);
```
matthewblain, line 227: also check a decrease?

server/appengine/src/main/java/who/RefreshCaseStatsServlet.java
```diff
@@ -317,11 +319,33 @@ protected void doGet(
       processWhoStats(rows, countryData, globalData);
     }
 
+    CaseStats oldCaseStats = StoredCaseStats.load(JurisdictionType.GLOBAL, "");
```
matthewblain, line 322: fixPartialLastDayAll should go here. That is, not inside the fetch loop. But before the new delta check.

matthewblain (commented):
noticed an issue with the previous change. the last result is only available after any pagination.

server/appengine/src/test/java/who/RefreshCaseStatsServletTest.java
brunobowden: Added decrease check

server/appengine/src/main/java/who/RefreshCaseStatsServlet.java
brunobowden: Good call. Fixed.

matthewblain (approved):

PR #1893, opened 2020-12-28
Last Updated Text

theswerd:
## Screenshots

<details>

</details>

- [x] iOS Simulator
- [ ] iOS Device
- [ ] Android Simulator
- [ ] Android Device
- [ ] other, please describe

## Checklist:

### DIFF
--- a/client/lib/components/recent_numbers_graph.dart
+++ b/client/lib/components/recent_numbers_graph.dart
@@ -269,12 +269,11 @@ class _RecentNumbersGraphState extends State<RecentNumbersGraph> {
     if (epochsMilliseconds == null) {
       return '';
     } else {
-      return 'Last Updated ' +
-          DateFormat.yMMMd().format(
-            DateTime.fromMillisecondsSinceEpoch(
-              epochsMilliseconds,
-            ),
-          );
+      return DateFormat.yMMMd().format(
+        DateTime.fromMillisecondsSinceEpoch(
+          epochsMilliseconds,
+        ),
+      );
     }
   }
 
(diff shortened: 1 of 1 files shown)


client/lib/components/recent_numbers_graph.dart
```diff
@@ -255,6 +264,20 @@ class _RecentNumbersGraphState extends State<RecentNumbersGraph> {
     }
   }
 
+  String get lastUpdated {
+    var epochsMilliseconds = widget?.timeseries?.last?.epochMsec?.toInt();
+    if (epochsMilliseconds == null) {
+      return '';
+    } else {
+      return 'Last Updated ' +
+          DateFormat.yMMMd().format(
```
matthewblain, line 273: how will this (eventually) get localized?

theswerd: We are probably going to have to consult an expert, on option is [the Dart **timeago** package](https://pub.dev/packages/timeago)

matthewblain: Looks like Flutter has all the tools expected for this, but the documentation is currently mangled. https://flutter.dev/docs/development/accessibility-and-localization/internationalization

I guess we'll address later for now. Not sure if we should put in some sort of comment marking known i18n/l10n issues in the code.

brunobowden: I think stylistically, it would be better to drop the "Last updated" and just have the date. There's also the benefit of avoiding a localization need but I'm not thinking about that so much.

brunobowden (approved):
Just drop the "Last updated"

client/lib/components/recent_numbers_graph.dart
matthewblain: Note that solves one problem, but does not that, say, en_GB uses 28 Dec 2020 while en_US uses Dec 28, 2020. But the good news is that dateformat does handle this case, though it needs more plumbing to make it work. I think this can probably just go in as is and will be very easy to find when the time comes.

Issue #1859, opened 2020-12-20
Text Alignment on Q&A Screen needs to get adjusted 


Mohit020888:
1. Launch the app
2. Complete Onboarding journey.
3. Tap on Learn tab in the footer.
4. Tap on Questions & Answers 
5. Expand the section which says "What should I do if I have been exposed to someone who has COVID-19?"

Expected Behavior
Text Alignment should proper

Actual Behavior 
Text Alignment is not proper.

System

- Device: Huawei Honor 7c
- OS: Android 8.0
- App Version: version 1.0.0-staging(281)

 Screenshots

![Screenshot_20201220-191403](https://user-images.githubusercontent.com/63721227/102721187-4d02ac00-431f-11eb-82d7-6acb8a744cae.jpg)

Checklist:

- [ ] Searched the [existing issues](https://github.com/WorldHealthOrganization/app/issues) to ensure you are not creating a duplicate.
- [ ] Followed the [Contributor Guidelines](https://github.com/WorldHealthOrganization/app/blob/master/docs/CONTRIBUTING.md).

champ96k:
@Mohit020888 and @brunobowden can please assign me this issue I would love to work on this issue :)

Thanks

Mohit020888:
@champ96k - thanks for volunteering to fix this one as well !
I have assigned it to you 👍

brunobowden:
@Mohit020888 @champ96k - it's going to be important that this looks good on very small screens. So please post a screenshot as part of any PR. Perhaps fixing the indentation will look better.... but also reduce the margin to the left of the bullet points as well.

My concern in part is that "unordered lists" or `<ul>` exist throughout the app content. So this should be a general fix for everything and not just this single page. See search results for all the content examples:
https://github.com/WorldHealthOrganization/app/search?q=%3Cul%3E

champ96k:
@brunobowden, @Mohit020888 Thanks for the detailed explanation, I definitely take care of all things

brunobowden:
@champ96k - main thing is that if the change starts looking too complex, then I'd question whether it's worth doing anything at all. It may be that the additional complexity is brittle and hard to maintain, which doesn't make sense for the product. See what you experiment with and feel free to share some preliminary before it takes up too much of your time.

champ96k:
[...]

PR #1912, opened 2021-01-06
solve Indentation on bullet points #1859 Issue

champ96k:
this issue solve suing [flutter_widget_from_html](https://pub.dev/packages/flutter_widget_from_html) 

System

- [x] OS: Android 9.0
- [x ] Device: Redmi 8A

### Screenshot:

Screenshot

## Checklist:

- [ x] Followed the [Contributor Guidelines](https://github.com/WorldHealthOrganization/app/blob/master/docs/CONTRIBUTING.md) and verified that all contributions are properly licensed pursuant to the [LICENSE](https://github.com/WorldHealthOrganization/app/blob/master/LICENSE).

### DIFF
--- a/client/lib/main.dart
+++ b/client/lib/main.dart
@@ -1,5 +1,4 @@
 import 'dart:async';
-
 import 'package:flutter/material.dart';
 import 'package:firebase_analytics/firebase_analytics.dart';
 import 'package:firebase_analytics/observer.dart';
--- a/client/lib/pages/question_index.dart
+++ b/client/lib/pages/question_index.dart
@@ -6,12 +6,12 @@ import 'package:who_app/api/linking.dart';
 import 'package:who_app/components/content_widget.dart';
 import 'package:who_app/components/loading_indicator.dart';
 import 'package:who_app/components/page_scaffold/page_scaffold.dart';
-import 'package:who_app/constants.dart';
 import 'package:flutter/cupertino.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter_html/flutter_html.dart';
-import 'package:html/dom.dart' as dom;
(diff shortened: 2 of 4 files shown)


champ96k:
Created new PR #1912

champ96k (commented):
self code review

brunobowden:
@champ96k - this is a creative use of the `flutter_widget_from_html` but the very thing I was concerned about is in the commit. There's a massive increase in the number of libraries, all of which need to go through some kind of security vetting and that's beyond what the volunteer team can manage right now. For example, I'm suspicious about why it pulls in a dependency like [wakelock](https://pub.dev/packages/wakelock)... which the WHO app doesn't need for any reason. We don't have the engineering capacity right now to consider such an extensive change. Adding these dependencies might well be a good move for other benefits, e.g. embedding videos and other functionality.... but this would be a large amount of work to just fix the indentation and the other benefits aren't clear at the moment. It's just not worth it at this point. See the earlier comment that was trying to speak to this: https://github.com/WorldHealthOrganization/app/issues/1859#issuecomment-753515800

P.S. You should include in the description "Fixes #1859" to have GitHub automatically close the issue if the PR was accepted. At this point, we won't be able to do that.

theswerd:
What is the benefit to using flutter_widget_from_html over flutter_html?

From what I have seen flutter_html is more generally used.

theswerd:
This also adds the open_iconic_flutter full icon pack.

champ96k:
> What is the benefit to using flutter_widget_from_html over flutter_html?
> 
> From what I have seen flutter_html is more generally used.

flutter_widget_from_html is a extension of flutter_html package with include some extra fetures

PR #1921, opened 2021-01-11
Match WHO Dashboard - Add one day to stats

### LINKED ISSUE #1917: Stats associate numbers with wrong day
## Summary

Scroll to the last day of a country stats. Note that it's associate with the wrong date - it's off by one day (can't remember if it's a day late or day early). This needs to match the https://covid19.who.int dashboard.

## Checklist:

theswerd:
Closes #1917

<details>

Simulator Screen Shot - iPhone 12 Pro Max - 2021-01-10 at 21 47 20
</details>

- [x] iOS Simulator
- [ ] iOS Device
- [ ] Android Simulator
- [ ] Android Device
- [ ] other, please describe

## Checklist:

### DIFF
--- a/client/lib/components/recent_numbers_graph.dart
+++ b/client/lib/components/recent_numbers_graph.dart
@@ -75,7 +75,13 @@ class _RecentNumbersBarGraphState extends State<RecentNumbersBarGraph> {
               final date = DateTime.fromMillisecondsSinceEpoch(
                   widget.timeseries[index].epochMsec.toInt());
               // Abbr to fit on single line: e.g. "Oct 18, 2020"
-              final formattedDate = DateFormat.yMMMd().format(date);
+              final formattedDate = DateFormat.yMMMd().format(
+                date.add(
+                  Duration(
+                    days: 1,
+                  ),
+                ),
+              );
 
               final formattedCount = NumberFormat().format(barRodData.y);
 
@@ -272,6 +278,10 @@ class _RecentNumbersGraphState extends State<RecentNumbersGraph> {
       return DateFormat.yMMMd().format(
         DateTime.fromMillisecondsSinceEpoch(
           epochsMilliseconds,
+        ).add(
+          Duration(
+            days: 1,
+          ),
         ),
       );
     }


matthewblain:
Is this a timezone issue, a server issue, or a parsing issue?

brunobowden:
@matthewblain - timestamps are for midnight, so it's ambiguous which day it applies to. It looks as though the WHO dashboard uses the date of when the data was reported to WHO - this is different to the date on which the case or death occurred. Adding one day is consistent with this, so the data that appears suddenly gets the date on which it was reported.

client/lib/components/recent_numbers_graph.dart
```diff
@@ -75,7 +75,13 @@ class _RecentNumbersBarGraphState extends State<RecentNumbersBarGraph> {
               final date = DateTime.fromMillisecondsSinceEpoch(
                   widget.timeseries[index].epochMsec.toInt());
               // Abbr to fit on single line: e.g. "Oct 18, 2020"
-              final formattedDate = DateFormat.yMMMd().format(date);
+              final formattedDate = DateFormat.yMMMd().format(
```
brunobowden, line 78: Extract this to a shared method to avoid duplicated code.

This should also be heavily commented to explain why it exists.

brunobowden (changes requested):

brunobowden:
@matthewblain - you could make a case for fixing this on the server.... but that would break the data consistency between the ArcGIS output and the rest API.

matthewblain:
I guess I think it should just be defined what it is. It's a time_t variant--are they all actual times, or midnight in some timezone, I looked at a sample of data from last month and they all appear to be midnight UTC.  DateTime.fromMillisecondsSinceEpoch will generate an item in the local timezone, though there is an option to have be UTC. Then formatdate does... something with it.

What is the date on the WHO dashboard generated from? Does it take the UTC midnight time and generate a string date from that, then display that everywhere in the world?

matthewblain:
As a concrete example, for Nigeria, on the WHO dashboard web site, it shows 1145 new cases for Dec 18. This matches with data for timestamp 1608249600000. Which is Dec 18, 0:00 Zulu. Which also happens to be 2020-12-17T16:00:00 in Pacific time.

Has this bug been verified with the device timezone set on both sides of UTC?

client/lib/api/who_date_format.dart
```diff
@@ -0,0 +1,12 @@
+import 'package:intl/intl.dart';
+
+extension WHOFormat on DateTime {
+  String get whoFormat => DateFormat.yMMMd().format(
+        // Timstamps are for midnight. It looks as though the WHO dashboard uses the date of when the data was reported to WHO rather than the original day.
```
brunobowden, line 5: The comment is missing the crucial information:
```
// Timestamps are for midnight so the day it applies to is ambiguous.
// Add 1 day to match the WHO Dashboard: http://covid19.who.int
```
The information about it being day reported rather than the day it happened should be a separate comment.

theswerd: @brunobowden changes made

matthewblain:
I just tested, and this is not the correct solution. It is a timezone issue.
Try this: Set the timezone on your device to, say, America/Los Angeles (currently UTC-8). It will show the date as being some particular day (in my case today, it shows the 24th). Change the timezone on your device to, say, Asia/Bangkok (currently UTC+7). Reload the stats page, and it will show the date as the next day (e.g. the 25th).

theswerd:
@matthewblain, do you think setting the timezone to UTC will work?

matthewblain:
Yes, it's quite possible using the isUtc argument on the constructor will do the trick:
DateTime.fromMillisecondsSinceEpoch(

    int millisecondsSinceEpoch,
    {bool isUtc: false}

)

brunobowden:
@matthewblain - I tip my hat to you. That would make more sense. Certainly, it shouldn't change depending on the local device time or time zone.

brunobowden:
I would've preferred that the server API used dates instead of timestamps to avoid any ambiguity.... but that's not important for now.

brunobowden (changes requested):
@theswerd - please look in to @matthewblain's comment about time zone. That is a plausible explanation to me. I don't know what time zone the data is in but I'll ask WHO.

client/lib/api/who_date_format.dart
```diff
@@ -0,0 +1,13 @@
+import 'package:intl/intl.dart';
+
+extension WHOFormat on DateTime {
```
brunobowden, line 3: I guess you wanted to write an extension ;-) I would've included the `DateTime.fromMillisecondsSinceEpoch` step also as that's common to both usages.... and it's the msSinceEpoch that has midnight ambiguity. By the time it's converted to a date, it should be correct, so doing +1 in the format step doesn't fit so well to me.

I would've just extracted a single method and not bothered with a new file. We can move it to a separate file if something else uses it. I think though that this formatting may be specific to the WHO dashboard.... no idea if it extends to the 

```
String whoDashboardDateFormat(int epochMS) {
  var date = DateTime.fromMillisecondsSinceEpoch(epochMS);
  // Timestamps are for midnight so the day it applies to is ambiguous.
  // Add 1 day to match the WHO Dashboard: http://covid19.who.int
  date.add(Duration(days: 1))
  // Abbr to fit on single line: e.g. "Oct 18, 2020"
  return DateFormat.yMMMd().format(date);
}
```

brunobowden (changes requested):

PR #1289, opened 2020-05-13
Display condition logic and new DSL features: country and rudimentary rollout/experimentation

advayDev1:
Display condition logic and new DSL features:
```
ctx_isoCountryCode == 'US' && inExperiment('13,54,93') - Shows something to 3 pct of US folks
ctx_localDays % 7 == 4 - Shows something once every week in user's local time
```

Implemented in Home, Learn, and Symptom Checker

Closes #1210 ; Closes #1211 ; Closes #1213 ; Affects #1209 ; Closes #1291 

TESTED: Android sim on Home/Learn/S-C with test yamls from: https://github.com/WorldHealthOrganization/app/commit/f4f2f124c3a33d1a403aa19d8558f48ae6f8f826

PR #1935, opened 2021-02-06
Restore symptom checker

patniemeyer:
This replaces the health check "poster" tab on the home page with the restored symptom checker code from commit: d614c406c and does what is necessary to get it working within the current app.   

All lint issues have been fixed and I believe this is working, not withstanding one known issue: that our application styled button does not show the disabled state.  I will open an issue for this separately.

Additionally the updated code does not utilize the observable pattern and possibly other app conventions adopted after it was written.  We can open issues for those going forward.

I have tested this in the simulator.


theswerd (approved):
LGTM

yukuairoy (approved):

client/lib/api/symptom_logic.dart
```diff
@@ -0,0 +1,62 @@
+import 'package:flutter/material.dart';
```
brunobowden, line 1: This should go in to the `client/flutter/lib/api/display_conditions.dart` file. It fits it with the cohort experimentation logic.

@advayDev1 moved the logic to `display_conditions.dart` in 1289:
https://github.com/WorldHealthOrganization/app/pull/1289/files#diff-773a1ae5f46e15d88eaf0fa31082951d904234896ec9456c8b5cef9ebb292056)

The removed it with 1476:
https://github.com/WorldHealthOrganization/app/pull/1476/files#diff-773a1ae5f46e15d88eaf0fa31082951d904234896ec9456c8b5cef9ebb292056

client/lib/pages/main_pages/check_up_poster_page.dart
```diff
@@ -12,6 +12,7 @@ import 'package:who_app/components/page_scaffold/page_scaffold.dart';
 import 'package:who_app/components/themed_text.dart';
 import 'package:who_app/constants.dart';
 
+// TODO: Remove if unused.
```
brunobowden, line 15: Remove this comment as we'll be keeping the `check_up_poster` around for certain countries where we can't get a check-up tool approved.

client/lib/pages/main_pages/app_tab_router.dart
```diff
@@ -7,7 +7,7 @@ import 'package:who_app/api/stats_store.dart';
 import 'package:who_app/api/user_preferences_store.dart';
 import 'package:who_app/api/who_service.dart';
 import 'package:who_app/constants.dart';
-import 'package:who_app/pages/main_pages/check_up_poster_page.dart';
+import 'package:who_app/pages/main_pages/check_up_intro_page.dart';
```
brunobowden, line 10: Please add the `import ....check_up_poster_page.dart';` as some versions of the app will use the poster where we can't get the check-up tool approved. So the client should import both packages and then allow toggling through the yaml file.

Can you test that switching the yaml changes what's displayed? Then confirm that in a PR comment.

client/lib/pages/main_pages/check_up_intro_page.dart
```diff
@@ -0,0 +1,166 @@
...
+          padding: EdgeInsets.zero,
+          child: ThemedText(
+            'See terms ›',
+            variant: TypographyVariant.body,
+            style: TextStyle(
+              color: Constants.whoBackgroundBlueColor,
+            ),
+          ),
+          onPressed: () {
+            return launch('https://whoapp.org/terms');
```
brunobowden, line 91: WRONG LINK!!! ;-)

Please use this logic instead:
https://github.com/WorldHealthOrganization/app/blob/8bb78d7f9c30d75805e4b1bdcf3cac3607bd6a33/client/lib/pages/about_page.dart#L33-L37

client/lib/pages/main_pages/routes.dart
```diff
@@ -35,9 +37,12 @@ class Routes {
           dataSource: Provider.of<ContentStore>(context),
           title: S.of(context).homePagePageButtonQuestions,
         ),
-    '/symptom-checker': (context) => CheckUpPosterPage(
+    '/symptom-checker-survey': (context) => SymptomCheckerView(
```
brunobowden, line 40: Does this need to preserve the `/symptom_checker` route so that it can be toggled in to use?

client/lib/pages/symptom_checker/symptom_checker_model.dart
```diff
@@ -0,0 +1,133 @@
+import 'package:flutter/material.dart';
+import 'package:who_app/api/content/schema/symptom_checker_content.dart';
+import 'package:who_app/api/display_conditions.dart';
+import 'package:who_app/api/symptom_logic.dart';
```
brunobowden, line 4: `import '....symptom_logic.dart';` wasn't needed before, is it needed now?

client/lib/pages/symptom_checker/symptom_checker_view.dart
```diff
@@ -0,0 +1,199 @@
...
+        ),
+      ),
+    );
+  }
+
+  Widget _buildPage(BuildContext context) {
+    if (_model == null) {
+      return _buildMessage('Loading...', loading: true);
+    }
+    if (_model.isFatalError) {
```
brunobowden, line 114: Errors like this should be logged:

https://github.com/WorldHealthOrganization/app/blob/aef8359dcddd2b213af8aa1aaf7c4ab7ea704531/client/lib/components/carousel/carousel.dart#L95

@theswerd - is there a way that `FirebaseAnalytics` can log an error rather than just an event? I want it to explicitly be an error so that it's not lost.

client/lib/pages/symptom_checker/symptom_checker_view.dart
```diff
@@ -0,0 +1,199 @@
...
+        children: _pages ?? []);
+  }
+
+  Widget _viewForPageModel(SymptomCheckerPageModel model) {
+    switch (model.question.type) {
+      case SymptomCheckerQuestionType.SingleSelection:
+        return ShortListQuestionView(pageDelegate: this, pageModel: model);
+      case SymptomCheckerQuestionType.MultipleSelection:
+        return ShortListQuestionView(pageDelegate: this, pageModel: model);
+    }
```
brunobowden, line 131: FirebaseAnalytics should log here too.... this might be caught by the top level main method:
https://github.com/WorldHealthOrganization/app/blob/5d4f17f49cbcfbea0c2badbdff0e35d6f278eb36/client/lib/main.dart#L83-L94

brunobowden (changes requested):
@patniemeyer - great to get this running again. Main comment is that we'll want to keep around the symptom poster for countries where we can't get the check up tool approved. We should import both of them in to the app and then allow toggling between them using the yaml config.

My concern is that this may require additional work so that the nav bar icon for "Check-Up" can be controlled by the `home_index.en.yaml` file and be configured. Since that's the case, let's deal with that in another PR and just focus on fixing up the other comments before landing this one.

client/assets/content_bundles/symptom_checker.en.yaml
```diff
@@ -0,0 +1,379 @@
+schema_version: 1
```
advayDev1, line 1: @brunobowden is this the content created entirely by WHO/team members, or is this the wording made earlier on....?

brunobowden: @advayDev1 - this is the same content as was removed in #1476. I've asked @patniemeyer to restore it as it was so we can check that it all works. There will be a separate process with WHO / Nigeria to figure out what the policy and text will be, likely that'll be something different from what it currently is.

advayDev1: See slack.  This was removed for a reason.

client/lib/pages/symptom_checker/symptom_checker_view.dart
patniemeyer: I'm not sure if we are curating a list of event names.  I'm adding one here for this:  `SymptomCheckerModelError`.

client/lib/pages/symptom_checker/symptom_checker_view.dart
patniemeyer: Maybe these should be the same error with different parameters?  I don't want to introduce more complexity to the event logs just for this so I'm going to call this `SymptomCheckerModelError2` for now.

client/assets/content_bundles/symptom_checker.en.yaml
patniemeyer: I'm not sure what to do about this so I'm going to leave it in the PR for now.

client/lib/pages/main_pages/app_tab_router.dart
patniemeyer: I'm adding a controls section to the content bundles and using it in the home_index.yaml like this: ```controls:
    - show_symptom_checker: true```

client/lib/pages/symptom_checker/symptom_checker_view.dart
brunobowden: Error2 is fine for now, main thing is to know an error is happening

brunobowden:
@patniemeyer - this is great. My one thought was that `show_symptom_checker` would be better as a property of the `symptom_checker.en.yaml` file and shouldn't be something that the top level app needs to know about. So the idea was that the main page would link to the Check-Up tool and then based on the `symptom_checker.en.yaml` config, it would either show the symptom page or the symptom checker tool. That has better encapsulation but I'm wary about it being a lot of refactoring work.

patniemeyer:
@brunobowden  Yes, that makes sense.  I will move it.

brunobowden:
Thanks Pat

Pat - ideally it’s generic part of the Symptom Checker like a question or a
results page. So if you wanted the power, you’d omit all the other sections
and just put in the content for the poster. That might be a bit much but go
as far as you can.

client/lib/api/content/content_bundle.dart
```diff
@@ -120,3 +137,19 @@ class ContentBundleSchemaVersionException implements Exception {}
...
+/// An open-ended map of control values that can be embedded into a content bundle.
+class ContentControls {
+  Map<String, String> map;
+
+  ContentControls(this.map);
+
+  String value(Object key, String defaultValue) {
+    return map[key.toString()] ?? defaultValue;
+  }
+
```
advayDev1, line 150: You might want to use the general display condition functionality that exists in the codebase for this.

patniemeyer:
As suggested I have merged the two pages into one CheckUpPage driven by the symptom_checker yaml.  I have tried to follow the ContentWidget pattern found in the app here but something is wrong: It seems that if the widget is built while content is null there is nothing to trigger a rebuild and the page remains empty until something down the line triggers a refresh.  It looks like the mobx observable stuff is supposed to cause the content widget to rebuild automatically... but this isn't working for some reason? I'm wondering if any of you are familiar with this code and see something obvious that I'm doing wrong...

client/assets/content_bundles/symptom_poster.en.yaml
```diff
@@ -1,39 +0,0 @@
-schema_version: 1
```
advayDev1, line 1: Can't delete since old apps rely on this url/file specifically.

advayDev1, line 1: also new schema version so old apps that see this will put up a force update message

brunobowden: Good point. What's the way to add a comment so we delete it later? Or maybe we should just track the delete with a GitHub issue.

patniemeyer: I have restored it.  We really only need this file on the server now though... We should have a place to archive them.

patniemeyer: The show upgrade dialog doesn't seem to be hooked up... I think it may have gotten lost at some point.

brunobowden (approved):

PR #1979, opened 2021-04-03
feat(v2 backend): updated the hack flavor to use v2 endpoints in the client

### LINKED ISSUE #1971: v2 backend: create client flavor that uses v2 backend
## Summary

Currently, the v2 backends aren't used by our clients (Android, iOS) yet.

## Desired Solution

To begin using the v2 backends, the easiest approach might be to introduce a "flavor" for each v2 backend that we'd like to connect to.

## Additional Context

The v2 backends currently support the v1 API, so should be relatively straightforward to use in the existing app. However, unlike in v1 backends, for v2 the URLs at which to access the dynamic API endpoints are different from URLs that serve the static content. More details are available in the `server`'s [README](https://github.com/WorldHealthOrganization/app/blob/master/server/README.md).

bezfeng:
This PR updates the client hack flavor to point the new V2 endpoints. The main changes are updating the `Endpoints` object to support the new URLs and updating the Firebase config files (in concert with registering the apps in the Firebase console).

Closes #1971

- [X] iOS Simulator
- [ ] iOS Device
- [X] Android Simulator
- [ ] Android Device
- [ ] other, please describe

## Checklist:


client/lib/api/endpoints.dart
```diff
@@ -2,34 +2,49 @@
...
           "Project: $projectId doesn't match prefix: $_whoMhPrefix");
     }
-    return projectId.substring(_whoMhPrefix.length);
+    return _isProd(projectId)
+        ? projectId.substring(_whoMhPrefix.length)
+        : projectId.substring(_whoMh2Prefix.length);
   }
 
   static String _serviceUrl(String projectId) {
+    if (_isProd(projectId)) {
```
brunobowden, line 27: Don't worry about making a breaking change here. You can remove the old prod code and we'll just get it all working once the production project is exposed publicly.

client/lib/api/endpoints.dart
```diff
@@ -2,34 +2,49 @@
 
 class Endpoint {
   static const _whoMhPrefix = 'who-mh-';
+  static const _whoMh2Prefix = 'who-mh2-';
   static const _prodProjectId = 'who-mh-prod';
   static const _prodServiceUrl = 'https://covid19app.who.int';
 
+  static const _region = 'europe-west6';
```
brunobowden, line 9: I'm a little wary that this parameter is repeated in the Terraform and Typescript code rather than being a shared parameter.

@rjhuijsman - is this something we can just configure through DNS? So the client connects to a custom domain and then we just point that domain to the right location? You mentioned some limitations on custom domains but I can't remember what they were.

brunobowden (commented):
@bezfeng - looking good but please connect with @rjhuijsman about the custom domain. Ideally the client shouldn't know anything about europe-west6 but instead it's all configured through DNS.

brunobowden (approved):
@bezfeng - we'll deal with the region thing later on. Freeing you up to make progress.

client/lib/api/endpoints.dart
bezfeng: Sounds good, removed the prod endpoints for now and left a TODO above `isProd` to get this working once V2 is deployed.

client/lib/api/endpoints.dart
bezfeng: Great point, agreed! Per discussion on Slack, I've added a TODO here to revisit hard coding the region in the client code.

PR #1981, opened 2021-04-07
feat(v2 backend): update FCM token directly in Firestore

### LINKED ISSUE #1973: v2 backend: remove or replace `putClientSettings` endpoint
## Summary

Currently, the `putClientSettings` endpoint that's part of the `v1` API is available in the `v2` backend as a HTTPS Cloud Function. However, that doesn't track with how we'd like to client-backend interaction; we'd like to move towards direct use of Firestore by the client, without a custom API layer in between.

In addition, the FCM token being updated through this call (its reason for existing) isn't currently being used - we should discuss the future of this functionality.

## Desired Solution

We'd like to:
1. Discuss whether the FCM token logic has a reason to exist in the project, or whether we should remove this functionality entirely. If the FCM token won't have a use at all, we should remove this functionality instead of doing the things below. If the FCM token _does_ have a use...
2. Replace the `putClientSettings` API with direct Firestore access.
3. Implement the `clientSettingsUpdated` Cloud Function with the logic to use the FCM token as desired.

## Additional Context

Cloud Functions live in `app/server/functions/src/index.ts`.

To allow for direct access to Firestore, we will need to update our `firestore.rules`. Please also update the unit tests in `firestore_rules.spec.ts` to show what clients should and should not be allowed to do under the new rules.

rjhuijsman:
Before this PR, we would update the client's FCM token by writing to the `putClientSettings` HTTPS endpoint. With this PR, we remove that intermediate API layer and write directly to Firestore.

We don't currently use the FCM token, so the main purposes of this PR are to...
1. Demonstrate how we'll perform direct Firestore access.
2. Demonstrate how to configure the app to use the local Firebase (Firestore) emulator.
3. Prepare the application for later use of the FCM token.

Closes #1973

## Screenshots

N/A

- [ ] iOS Simulator
- [ ] iOS Device
- [X] Android Simulator
- [ ] Android Device
- [X] other, please describe: updated and ran unit tests in both `client` and `server`.

## Checklist:

### DIFF
--- a/client/README.md
+++ b/client/README.md
@@ -66,6 +66,8 @@ Then, in your `client` directory, update your `main.dart` as follows:
 
     const USE_FIREBASE_LOCAL_EMULATORS = true;
 
+When working with the iOS client, temporarily disable transport security [as documented here](https://firebase.flutter.dev/docs/installation/ios/#enabling-use-of-firebase-emulator-suite).
+
 Then, run your application in its `hack` flavor:
 
     flutter run --flavor=hack
--- a/client/README.md
+++ b/client/README.md
@@ -53,3 +53,23 @@ For example to access the `staging` server (the default), the files are:
 Android: client/android/app/src/<flavor>/google-services.json
 iOS:     client/ios/config/<flavor>/GoogleService-Info.plist
 ```
+
+##### Using the Firebase Emulators
+
+If you'd like to test your app without using a "real" Firebase project, you can use the Firebase Local Emulator Suite as follows.
+
+First, start your local emulators by navigating to your `server/functions` directory and running:
+
+    firebase emulators:start --project=dev
+
+Then, in your `client` directory, update your `main.dart` as follows:
+
+    const USE_FIREBASE_LOCAL_EMULATORS = true;
+
+Then, run your application in its `hack` flavor:
+
+    flutter run --flavor=hack
+
+Your logs will confirm that you are using the local emulators:
+
+    I/flutter (13491): Will use local 🔥🔥 Firebase 🔥🔥 emulator
--- a/client/lib/api/who_service.dart
+++ b/client/lib/api/who_service.dart
@@ -7,12 +7,18 @@ import 'dart:io' as io;
 
 import 'package:who_app/proto/api/who/who.pb.dart';
 
+import 'package:cloud_firestore/cloud_firestore.dart';
 import 'package:firebase_performance/firebase_performance.dart';
+
+const CLIENT_COLLECTION = 'Client';
 
 class WhoService {
   final String serviceUrl;
+  final FirebaseFirestore firestore;
 
-  WhoService({@required String endpoint}) : serviceUrl = endpoint;
+  WhoService({@required String endpoint})
+      : serviceUrl = endpoint,
+        firestore = FirebaseFirestore.instance;
 
   final _MetricHttpClient http = _MetricHttpClient(
     Client(),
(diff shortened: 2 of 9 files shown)


client/README.md
```diff
@@ -53,3 +53,23 @@ For example to access the `staging` server (the default), the files are:
...
+
+    firebase emulators:start --project=dev
+
+Then, in your `client` directory, update your `main.dart` as follows:
+
+    const USE_FIREBASE_LOCAL_EMULATORS = true;
+
+Then, run your application in its `hack` flavor:
+
+    flutter run --flavor=hack
```
brunobowden, line 71: Why the `hack` flavor? That's really meant for penetration testing. I was expecting people to use the staging flavor.... which I think it should default to anyway.

client/lib/api/who_service.dart
```diff
@@ -7,32 +7,36 @@ import 'dart:io' as io;
...
-    if (response.statusCode != 200) {
-      throw Exception('Error status code: ${response.statusCode}');
+    var clientId = await UserPreferences().getClientUuid();
+    try {
+      await firestore.collection(CLIENT_COLLECTION).doc(clientId).set({
+        'uuid': clientId,
+        'token': token,
+        'disableNotifications': token == null || token.isEmpty,
+        'platform': _platform,
+        'isoCountryCode': isoCountryCode
```
brunobowden, line 36: That's much more concise

client/lib/main.dart
```diff
@@ -37,6 +38,10 @@ PackageInfo _packageInfo;
 
 PackageInfo get packageInfo => _packageInfo;
 
+// ATTENTION: never check this in as 'true'! Always set back to 'false' before
+//            sending out your PR. This is verified by `test/firebase_test.dart`.
```
brunobowden, line 42: `test/firebase_test.dart` looks like a file but I don't see it in this PR? Good to have a test for mistakes like this.

server/firestore.rules
```diff
@@ -1,10 +1,21 @@
 rules_version = '2';
 service cloud.firestore {
   match /databases/{database}/documents {
-    // Currently: do not allow any direct reads or writes.
-    // TODO: this will change as we update the client to do direct reads and writes.
+    // Currently: do not allow any direct reads or writes unless otherwise specified.
```
brunobowden, line 4: List this as the default instead?
```suggestion
    // Default: do not allow any direct reads or writes unless otherwise specified.
```

server/functions/src/index.ts
```diff
@@ -39,69 +39,6 @@ export const getCaseStats = functions
     response.status(200).json(data);
   });
 
-// Implementation of the v1 API"s `putClientSettings` method.
-// TODO: replace with direct Firestore acccess from the client.
-export const putClientSettings = functions
```
brunobowden, line 44: Wonderful to delete code.....

brunobowden (approved):
@rjhuijsman - I'm happy but good to have someone else with Firestore experience look it over

@theswerd - care to take a look?

client/README.md
rjhuijsman: Currently the only flavor using our `v2` backend is `hack` (thanks to Bez's #1971), but once the other endpoints are removed in favor of direct Firestore use we should be able to replace this with any flavor we want.

client/lib/main.dart
rjhuijsman: Oops! Added the missing file.

server/functions/src/index.ts
rjhuijsman: 🎉

client/lib/api/who_service.dart
rjhuijsman: 🎉

matthewblain (commented):
looks good. just more comment comments. 
i did tewak how putClinetSettings itself was called at some point not too long ago, as it relates to needing the country and the enable flags both being there.. but I don't think it's worth figuring out a way to optimize it further now.  rather, the next change is likely to be about language localization. and probably the same call sequence.

client/README.md
```diff
@@ -53,3 +53,23 @@ For example to access the `staging` server (the default), the files are:
...
+
+Your logs will confirm that you are using the local emulators:
+
+    I/flutter (13491): Will use local 🔥🔥 Firebase 🔥🔥 emulator
```
bezfeng, line 75: Might also be worth noting that for iOS, you may need to temporarily disable app transport security since it seems like the emulator sends traffic over http instead of https:

https://firebase.flutter.dev/docs/installation/ios/#enabling-use-of-firebase-emulator-suite

client/lib/api/who_service.dart
```diff
@@ -7,32 +7,36 @@ import 'dart:io' as io;
...
+      });
+    } catch (e) {
+      debugPrint('Failed to update FCM token in Firestore: $e');
```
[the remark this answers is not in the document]
bezfeng: Yup, I believe all the syncing logic is handled behind the scenes by Firestore now and it should automatically retry failed writes. In fact, there doesn't even seem to be a way for the client to manually request a sync.

bezfeng (approved):
Looks great!

client/README.md
rjhuijsman: Great catch, thank you!

PR #1986, opened 2021-04-22
Started v2.0 migration

theswerd:
In progress

## Checklist:


client/lib/api/observer_provider.dart
```diff
@@ -0,0 +1,79 @@
+import 'package:flutter/widgets.dart';
```
advayDev1: As requested, 0.0.2-nullsafety.0 published.  If you are going to copy this in instead, plz make sure to follow the license/noticing requirements.

brunobowden: @advayDev1 - thanks. Just to make sure we get this right. Would the following notice in the commit message be sufficient for you or should we do anything different?

```
client/lib/api/observer_provider.dart copied under Apache 2.0 license from this original source (copyright @advayDev1 - Advay Mengle):
https://github.com/holvonix-open/observer-provider/blob/master/LICENSE
```

brunobowden: @advayDev1 - thank you

Corrected Message below.... Just to make sure we get this right. Would the following notice in the commit message be sufficient for you or should we do anything different?

```
client/lib/api/observer_provider.dart copied under Apache 2.0 license from
this original source (copyright @advayDev1 - Advay Mengle):
https://github.com/holvonix-open/observer-provider/blob/master/lib/src/observer_provider.dart
```

advayDev1: It's just a standard "NOTICE" - the original NOTICE from the original file can be copied into the third-party license notices where all the other ones are.  The code is actually open sourced by Holvonix as a company, not me individually.

theswerd: This is a placeholder - I plan on deleting it now that your package is upgraded to null safety.

client/lib/components/home_page_sections/home_page_recent_numbers.dart
```diff
@@ -69,7 +69,7 @@ class __HomeStatsFaderState extends State<_HomeStatsFader>
     with SingleTickerProviderStateMixin {
   bool fadeState = true;
 
-  AnimationController _animationController;
+  late AnimationController _animationController;
```
brunobowden, line 72: late is a good syntax

client/pubspec.yaml
```diff
@@ -18,59 +18,59 @@ issue_tracker: https://github.com/WorldHealthOrganization/app/issues
...
+  mobx: ^2.0.1
+  notification_permissions: ^0.5.0
+  # observer_provider: 0.0.1+2
+  package_info: "2.0.0"
+  # page_view_indicator: ^2.0.4
+  path_drawing: ^0.5.0
+  path_provider: ^2.0.1
+  protobuf: ^2.0.0
+  provider: ^5.0.0
   # TODO: upgrade to ^0.6.5+2, build warning treated as error
```
brunobowden: Remove `#TODO` now that it's fixed

client/pubspec.yaml
```diff
@@ -18,59 +18,59 @@ issue_tracker: https://github.com/WorldHealthOrganization/app/issues
...
-  provider: ^4.3.2
+  flutter_cache_manager: 3.0.1
+  flutter_html: ^2.0.0-nullsafety.0
+  flutter_mobx: ^2.0.0
+  flutter_svg: ^0.21.0-nullsafety.0
+  font_awesome_flutter: ^9.0.0
+  http: ^0.13.1
+  mobx: ^2.0.1
+  notification_permissions: ^0.5.0
+  # observer_provider: 0.0.1+2
```
brunobowden: Adding this back again?

client/pubspec.yaml
```diff
@@ -18,59 +18,59 @@ issue_tracker: https://github.com/WorldHealthOrganization/app/issues
...
+  package_info: "2.0.0"
+  # page_view_indicator: ^2.0.4
```
brunobowden, line 52: Make this a TODO?

brunobowden (commented):
@theswerd - this looks good to me. What else does it need?

theswerd:
just looked over the cache manager - reimplementation should be relatively easy. they provided instructions for moving away from base cache manager.

brunobowden:
Thanks Ben. That's encouraging.

client/lib/api/observer_provider.dart
theswerd: Returned to @advayDev1's package

client/pubspec.yaml
theswerd: Done

theswerd:
Cache manager working - only need to re-implement charts

brunobowden (commented):
Changes to cache manager look good

All we need is `fl_chart` now...

client/lib/pages/onboarding/country_list_page.dart
```diff
@@ -45,7 +45,7 @@ class _CountryListPageState extends State<CountryListPage> {
           decoration: InputDecoration(
               border: InputBorder.none, hintText: 'Enter your country'),
           onChanged: (String value) async {
-            List filtered = (widget.countries?.values ?? []).where((element) {
+            List filtered = (widget.countries.values).where((element) {
```
brunobowden, line 48: This is much cleaner. Good case where null safety enforces better discipline

client/lib/components/home_page_sections/home_page_recent_numbers.dart
```diff
@@ -96,7 +96,9 @@ class __HomeStatsFaderState extends State<_HomeStatsFader>
   Widget build(BuildContext context) {
     final countryList = Provider.of<IsoCountryList>(context);
     final currentCountryCode = widget.statsStore.countryIsoCode;
-    final country = countryList.countries[currentCountryCode!];
+    var country;
+    if (currentCountryCode != null)
+      country = countryList.countries[currentCountryCode];
```
brunobowden, line 101: I was thinking you could simplify this with [Conditional Property Access](https://dart.dev/codelabs/dart-cheatsheet#conditional-property-access)... then I realize that doesn't work and you need to do it this way instead ;-)

brunobowden (approved):
@theswerd @bezfeng - great to get this working. I'll send another PR to update the build settings

client/lib/components/home_page_sections/home_page_recent_numbers.dart
bezfeng: Ah good call on simplifying this bit! After playing around some more, I think we can just do:

`final country = countryList.countries[currentCountryCode];`

(sorry, I could've sworn this was causing a runtime error but it's working fine for me now 😛)

Issue #1972, opened 2021-03-30
v2 backend: generate and serve case stats


rjhuijsman:
## Summary

Currently, the case stats being served by the v2 backend are...
1. Static, out-of-date, not-localized stats, useful only as an example.
1. Served by the old `v1` API (`getCaseStats`) instead of using Firestore-based "API-less" access.

## Desired Solution

We'd like to...
1. Determine a good data layout in which to store case stats in Firestore Native.
1. Periodically generate actual case stats (via [scheduled Cloud Function](https://firebase.google.com/docs/functions/schedule-functions))
1. Modify the clients to read the case stats directly from Firestore, instead of going through the `v1` API.

## Additional Context

There is existing AppEngine code to generate case stats in [`RefreshCaseStatsServlet.java`](https://github.com/WorldHealthOrganization/app/blob/master/server/appengine/src/main/java/who/RefreshCaseStatsServlet.java). We're looking to replace this code with our new Cloud Functions approach - its logic can likely be mostly maintained.

To allow for direct access from the client to Firestore, we will need to update our Firestore security rules in `firestore.rules`. Please also update the unit tests in `firestore_rules.spec.ts` to demonstrate which parts of the database should and should not be accessible directly to clients.

PR #1988, opened 2021-05-05
WIP: Refreshcasestats using Functions/Firestore

matthewblain:
This rewrites the refreshCaseStats functionality in Typescript on the Google Cloud Functions platform.

Current status is that it is functional: It can load the casestats from the WHO/ArcGIS endpoint and store them in the Datastore/Firestore. This functionality should be exactly the same as the old system. It has a getCaseStats endpoint which should return the data in the equivalent format as the old system.

What is missing?
 - Unit tests.
 - Automated 'cron' type functionality. This currently only runs on an (unsecured) http request endpoint.
 - Possibly changing how the data is stored in Firestore from a single entity (doc) per country to a collection per country. This would be to make syncing easier on the client, in case firestore sync works better that way.
 - Also, the code is totally non-idomatic Typescript, as this is the first time I've used TS. 

#1972 

## Screenshots

N/A

- [ ] iOS Simulator
- [ ] iOS Device
- [ ] Android Simulator
- [ ] Android Device
- [x] `curl` to a dev App Engine server
- [ ] other, please describe

Well, curl to the google cloud functions/firestore emulator.

## Checklist:


server/functions/package.json
```diff
@@ -3,10 +3,12 @@
...
     "build": "tsc",
-    "test": "firebase --project=$GCLOUD_PROJECT emulators:exec \"mocha -r ts-node/register src/**/*.spec.ts\"",
+    "test_rules": "firebase --project=mblain-13nov2020 emulators:exec \"mocha -r ts-node/register src/**/*.spec.ts\"",
+    "test": "mocha -r ts-node/register --reporter spec",
     "check-location": "./check_location.sh",
-    "serve": "npm run build && firebase emulators:start",
-    "shell": "npm run build && firebase functions:shell",
+    "preserve": "tsc",
+    "serve": "firebase emulators:start --project=mblain-13nov2020",
+    "shell": "npm run build np&& firebase functions:shell",
```
rjhuijsman, line 11: Was there a usability issue with the tests that you're working around here? I'm concerned that in this config just running `npm test` will not work unless you knew to previously start the emulators, but if the tests were hard for you to use prior to this change we should address that!

Also, we'll want to get rid of hardcoded project ID's before submitting this, obviously. :)

server/functions/src/index.ts
```diff
@@ -12,31 +13,114 @@ const COUNTRY_CODE = /^[A-Z][A-Z]$/; // Literal regex for immediate compilation.
...
   uuid: string;
   token: string;
   disableNotifcations: boolean;
-  platform: Platform;
+  platform: who.Platform;
   isoCountryCode: string;
   subscribedTopics: string[];
 }
 
+refreshcasestats.setDb(db); /// This seems wrong???????????????????????????????????????????????????????????
```
rjhuijsman, line 25: It's fine to run...

```
admin.initializeApp();
const db = admin.firestore();
```

... in every file where we need database access. Alternatively, we can create a `db.ts` which has those two lines and an `export db` that we can then import in all the other files.

server/functions/src/index.ts
```diff
@@ -1,7 +1,8 @@
 import * as admin from "firebase-admin";
 import * as functions from "firebase-functions";
-import * as fs from "fs";
 import { SERVING_REGION } from "./config";
+import * as refreshcasestats from "./refreshcasestats"; // Is this right?
```
rjhuijsman, line 4: This is fine! You might also consider importing only specific symbols (like we do from `./config` above), notably `refreshCaseStats`, so that the export (further down) can just be `export refreshCaseStats;` - but only once we can remove the `setDb()` call, since that would look very strange without the context of `refreshcasestats`.

server/functions/src/index.ts
```diff
@@ -12,31 +13,114 @@ const COUNTRY_CODE = /^[A-Z][A-Z]$/; // Literal regex for immediate compilation.
...
+    jurisdictionStats: [],
+    ttl: 0,
+  } as who.GetCaseStatsResponse;
+
+  for (let jurisdiction of jurisdictions) {
+    let key = `${who.jurisdictionTypeFromJSON(jurisdiction.jurisdictionType)}:${
+      jurisdiction.code
+    }`;
+    console.log(`Loading data for ${key}`);
+    let loadedData = await db.collection("StoredCaseStats").doc(key).get();
```
rjhuijsman: We'll probably want to make `"StoredCaseStats"` a constant loaded from `refreshcasestats.ts`?

server/functions/src/index.ts
```diff
@@ -12,31 +13,114 @@ const COUNTRY_CODE = /^[A-Z][A-Z]$/; // Literal regex for immediate compilation.
...
 export const getCaseStats = functions
   .region(SERVING_REGION)
   .https.onRequest((request, response) => {
-    // A "fake" of the v1 API: we're just going to return a static file that we
-    // happen to have on disk.
-    // TODO: replace this with actual statistics in Firestore.
-    const data = JSON.parse(fs.readFileSync("casestats.json", "utf8"));
-    response.status(200).json(data);
+    getCaseStatsAsync(request.body)
+      .then(function (resp) {
```
rjhuijsman: It's generally preferred to use the now-more-common shorthand for an inline function definition:
```
(params) => { body }
```
There's a few examples throughout this file.

server/functions/src/index.ts
```diff
@@ -12,31 +13,114 @@ const COUNTRY_CODE = /^[A-Z][A-Z]$/; // Literal regex for immediate compilation.
...
+        response.status(200).json(resp);
+      })
+      .catch(function (error) {
+        console.log(error);
+        response.status(400).send("Error"); // distinguish 400 and 500?
+        return;
+      });
```
rjhuijsman: You can use async/await here too (`async (request, response) => { ... }`) if you'd like.

server/functions/src/index.ts
```diff
@@ -12,31 +13,114 @@ const COUNTRY_CODE = /^[A-Z][A-Z]$/; // Literal regex for immediate compilation.
...
-    // TODO: replace this with actual statistics in Firestore.
-    const data = JSON.parse(fs.readFileSync("casestats.json", "utf8"));
-    response.status(200).json(data);
+    getCaseStatsAsync(request.body)
+      .then(function (resp) {
+        response.status(200).json(resp);
+      })
+      .catch(function (error) {
+        console.log(error);
+        response.status(400).send("Error"); // distinguish 400 and 500?
```
rjhuijsman: Normally, yes, I think it would be nice if we could distinguish between "I don't know that jurisdiction" and "I'm broken".

However, we're aiming to replace this code with direct-to-Firestore access soon - so I feel we can probably omit such niceties.

server/functions/src/refreshcasestats.ts
```diff
@@ -0,0 +1,371 @@
+import axios from "axios";
+import * as admin from "firebase-admin";
+import * as functions from "firebase-functions";
+import { SERVING_REGION } from "./config";
+import * as who from "./who";
+
+// THis seems wrong to initializeApp both here and index.ts?
+//admin.initializeApp();
+//const db = admin.firestore();
```
rjhuijsman, line 9: It's fine, common practice even, but perhaps most cleanly we could import `db` from a common `db.ts` that `export const db = admin.firestore()`.

server/functions/src/refreshcasestats.ts
```diff
@@ -0,0 +1,371 @@
...
+// The following line causes the Firebase emulator to say:
+//  "It appears your code is written in Typescript, which must be compiled before emulation."
+//const db = admin.firestore();
```
rjhuijsman, line 12: That's puzzling! I... Have no idea why it would say that. 😆

server/functions/src/refreshcasestats.ts
```diff
@@ -0,0 +1,371 @@
...
+// This crazy mess makes it work.
+
+var db: admin.firestore.Firestore;
+export function setDb(newDb: admin.firestore.Firestore) {
+  db = newDb;
+}
```
rjhuijsman, line 18: It's not thaaaat crazy - but I'd like the other options better.

server/functions/src/refreshcasestats.ts
```diff
@@ -0,0 +1,371 @@
...
+
+// Triggers to reject case stats update for unexpectedly large increase
```
rjhuijsman: Nit: "trigger" in the context of a Cloud Function is a thing that causes the Cloud Function to run. I'd prefer to not overload the name.

I'd describe this type of parameter as a "safety threshold" or (less common, but I like this term) a "fuse".

server/functions/src/refreshcasestats.ts
```diff
@@ -0,0 +1,371 @@
...
+  snapshot.totalDeaths += totalDeaths;
+}
+
+/**
+ * Process the JSON data returned by the ArcGIS server.
+ *
+ * @param features        JSON features from ArcGIS server
+ * @param countryData Accumulated data for each country.
+ * @param globalData  Accumulated data, sum of all countries.
+ */
```
rjhuijsman: Comments like this are best placed directly above the function definition.

server/functions/src/refreshcasestats.ts
```diff
@@ -0,0 +1,371 @@
...
+}
+
+interface Feature {
+  attributes: Attributes;
+}
+
+// The ARCGISResponse contains more than this, but all we need are the features.
+interface ArcGISResponse {
+  features: Feature[];
+}
```
rjhuijsman, line 118: I like that we have this here, very nice code-as-documentation.

server/functions/src/refreshcasestats.ts
```diff
@@ -0,0 +1,371 @@
...
+          response.status(500).send("Error");
+          return;
+        }
+        response.status(200).send("OK");
+      })
+      .catch(function (error) {
+        console.log(error);
+        response.status(500).send("Error"); // How to get the system to backoff/retry? What text to use?
+        return;
+      });
```
rjhuijsman: You can use async/await here too (`async (request, response) => { ... }`) if you'd like.

server/functions/src/refreshcasestats.ts
```diff
@@ -0,0 +1,371 @@
...
+  });
+
+// TODO: How do we specify which region this should run in? This data is public,
+// but we presumably want it to run in the same region as our datastore and
+// other code.
+// TODO: Scheduled functions require billing (lol). It's only $0.10/month but
+// but that's absurd for a tiny project...
```
rjhuijsman, line 365: It's not even the functions that require billing... It's the scheduler that fires off the Pub/Sub message that _triggers_ the function that needs it... Le sigh. 😕 

FWIW, even on billing-enabled projects the free tier is in effect, so we'll likely still be charged $0.

server/functions/src/who.ts
```diff
@@ -0,0 +1,144 @@
+/* eslint-disable */
+
+// This was generated from who.proto with https://github.com/stephenh/ts-proto
+// Then all of the encoding methods were removed as they interfered with
+// Firestore and are not necessary.
+// This leaves only the interface definitions.
```
rjhuijsman, line 6: I like this a lot! I in particular like how the use of these interfaces produces very nice self-documenting code in the other `.ts` files.

server/functions/src/index.ts
```diff
@@ -12,31 +13,114 @@ const COUNTRY_CODE = /^[A-Z][A-Z]$/; // Literal regex for immediate compilation.
...
+async function getCaseStatsAsync(req: who.GetCaseStatsRequest) {
+  let jurisdictions = req.jurisdictions;
+  if (jurisdictions == undefined) {
+    throw Error("Bad request.");
+  }
+  let resp = {
+    globalStats: undefined,
+    jurisdictionStats: [],
+    ttl: 0,
+  } as who.GetCaseStatsResponse;
```
rjhuijsman: Consider this slight change:
```
  let resp: who.GetCaseStatsResponse = {
    globalStats: undefined,
    jurisdictionStats: [],
    ttl: 0,
  }
```

... Since the `as` type assertion will _not_ throw a type error if you're missing required fields (I'm basing that on e.g. [this](https://basarat.gitbook.io/typescript/type-system/type-assertion#assertion-considered-harmful)), but when constructing an object of a given type you _will_ be warned about that.

server/functions/src/index.ts
matthewblain: This does not seem to be correct. If I call it in both files, I get an error like this:

`!  Error: The default Firebase app already exists. This means you called initializeApp() more than once without providing an app name as the second argument. In most cases you only need to call initializeApp() once. But if you do want to initialize multiple apps, pass a second argument to initializeApp() to give each app a unique name.`

server/functions/src/index.ts
matthewblain: That's the database schema question! What are the entity kind names, and where is that documented in the code? Is it shared 'client' and 'server'? For now, I added a const.

server/functions/src/index.ts
matthewblain: OK

server/functions/src/refreshcasestats.ts
matthewblain: Many of the comments in this file are copy/pasted from refreshcasestats.java.! Made some wording changes.

server/functions/src/refreshcasestats.ts
matthewblain: Hah. Also, what is the ts docstring format anyway? This is javadoc.

server/functions/src/who.ts
matthewblain: However it needs better tooling to keep in sync w/proto changes. Probably just some flags to ts-proto but I have not investigated.

server/functions/src/index.ts
matthewblain: thanks! i didn't know how to do 'constructors' but copied another example in this file.

server/functions/package.json
matthewblain: The current system assumes my shell is bash and some other things. My version is worse in that it hardcodes some stuff. I think there's some package.json way to do the dependecies (build -> serve). Interestingly now if I do npm run serve it does get it right. Test, not sure.

server/functions/src/index.ts
matthewblain: OK done.

server/functions/src/refreshcasestats.ts
matthewblain: See comment above. Maybe you can try/debug on your machine?

server/functions/src/refreshcasestats.ts
matthewblain: ok

matthewblain:
See comments.
Note: I have not tried to actually deploy! 
Note2: If the request to getcasestats is not valid json, the function returns immediately... but (at least the emulator) keeps the function running in the background for 60s then kills it. Strange, IIRC it's the opposite of app engine. (IIRC, on return, appengine kills the request? Or vice versa, the request doesn't return until it's done?).  Anyway I have no idea why this is, or if it matters. I saw the same behavior when I had some bugs in refreshcasestats (I fired off but did not wait for some async requests), but in that case at least it sort of makes sense.
Note3: Still not unit tests.

server/functions/src/index.ts
rjhuijsman: Oh shoot, that unfortunately makes a lot of sense.

In that case, I'd go for a separate `db.ts` module that we import from everywhere.

rjhuijsman:
> See comments.

> Note: I have not tried to actually deploy!

If the emulators work the deploy will most likely work too, but we'll of course have to verify before we submit.

> Note2: If the request to getcasestats is not valid json, the function returns immediately... but (at least the emulator) keeps the function running in the background for 60s then kills it. Strange, IIRC it's the opposite of app engine. (IIRC, on return, appengine kills the request? Or vice versa, the request doesn't return until it's done?). Anyway I have no idea why this is, or if it matters. I saw the same behavior when I had some bugs in refreshcasestats (I fired off but did not wait for some async requests), but in that case at least it sort of makes sense.

Interesting... Do you get the error 400 from `getCaseStats`, as one might expect?

The behavior you observe would be consistent with my expectations if the code failed to send an HTTP response before returning - but if there _is_ a response being sent (as I believe the `catch` in `getCaseStats` should ensure) then I believe you're likely seeing emulator behavior that's not consistent with production.

> Note3: Still not unit tests.

server/functions/src/refreshcasestats.ts
```diff
@@ -0,0 +1,371 @@
...
+      globalData,
+      attributes.date_epicrv,
+      attributes.NewDeath,
+      attributes.CumDeath,
+      attributes.NewCase,
+      attributes.CumCase
+    );
+
+    let isoCode = attributes["ISO_2_CODE"];
+    if (isoCode) {
```
rjhuijsman, line 144: Please make the expectation type-explicit: `if (isocode !== undefined) {`

Otherwise, the reader doesn't know whether you're expecting `undefined`, `null`, `0`, `"0"`, `""`, etc.

server/functions/src/refreshcasestats.ts
```diff
@@ -0,0 +1,371 @@
...
+
+      updateData(
+        countryData.get(isoCode)!,
+        attributes.date_epicrv,
+        attributes.NewDeath,
+        attributes.CumDeath,
+        attributes.NewCase,
+        attributes.CumCase
+      );
+    }
```
rjhuijsman, line 168: Given that this block has no `else`, consider:
```
if (isoCode === undefined) {
	continue
}
// Remainder of code at one level of indentation lower.
```

server/functions/src/refreshcasestats.ts
```diff
@@ -0,0 +1,371 @@
...
+  let globalData = {
+    jurisdictionType: who.JurisdictionType.GLOBAL,
+    jurisdiction: "",
+    lastUpdated: 0,
+    cases: 0,
+    deaths: 0,
+    recoveries: 0,
+    attribution: "WHO",
+    timeseries: [],
+  } as who.CaseStats;
```
rjhuijsman, line 186: Also remove `as` assertion here in favor of type annotation?

server/functions/src/refreshcasestats.ts
```diff
@@ -0,0 +1,371 @@
...
+
+    // Consider removing last day
+    if (lastSnapshot.dailyCases == 0 && lastSnapshot.dailyDeaths == 0) {
+      if (priorSnapshot.dailyCases > 0 || priorSnapshot.dailyDeaths > 0) {
+        // Likely "No Data Reported"
+        timeseries.pop();
+        countryStat.lastUpdated = priorSnapshot.epochMsec;
+        countriesLastDayRemoved.push(countryName);
+      }
+    }
```
rjhuijsman, line 257: Given that these `if`s don't have `else`s, consider using `if (opposite) continue` here too, to improve readability by reducing indentation.

rjhuijsman (changes requested):
No major code issues to my eyes! 

To me, the remaining work looks like:
1. Factor out the `firebase.initializeApp(); export db = firebase.firestore();` into a `db.ts` that can get imported by the other files.
2. Fix up the `package.json` so that it's easily usable (ideally no hardcoded project IDs, if that's not feasible then make the hardcoded values something like `YOUR-PROJECT-ID-HERE` and update `README.md` with instructions).
3. Unit tests where useful.
4. Remove the HTTP function and replace with the scheduled async function (note that scheduled functions can be triggered manually too by clicking on the Cloud Console's Cloud Scheduler page).
5. Deploy to a real project and verify there.

How does that sound?

server/functions/src/refreshcasestats.ts
rjhuijsman: I'd be very surprised if this issue persists after we move initialization to a separate `db.ts` module. Happy to help debug if it's still the case then.

server/functions/src/refreshcasestats.ts
matthewblain: I think having the exceptional case in an 'if' is more understandable.

server/functions/src/index.ts
matthewblain: Creating a new module seems to work. Not sure if I did it correctly.

